PR #42, opened 2024-09-12
Jobs

barbinbrad:


barbinbrad:
Landed in carbon PR #1540: ERP + MES now set baseline security headers on document responses — a **safe CSP subset** (`object-src 'none'; base-uri 'self'; frame-ancestors 'self'`) that controls mobile code without a restrictive `default-src`/`script-src` (won't break the SPA), plus nosniff / SAMEORIGIN / Referrer-Policy / HSTS. 3.13.13 → MET on merge.

PR #323, opened 2025-11-04
Force mount board scroll and make it bigger

kartikbajaj1:
image
image
scroll only shows when required, its little bigger than before should make is easy to access..


barbinbrad:
@kartikbajaj1 there's a similar component in `apps/mes` can you do the same change there?

PR #330, opened 2025-11-06
Adding measurement standards + attempts + some renamings

kartikbajaj1:
image


packages/database/supabase/migrations/20251106045123_view.sql
```diff
@@ -0,0 +1,9 @@
+DROP VIEW IF EXISTS "gaugeCalibrationRecords";
```
kartikbajaj1: didn't realize early that I needed to recreate the view, so had to add another migration

PR #331, opened 2025-11-07
Moved item count out of pagination to show for all tables across the app

kartikbajaj1:
image


barbinbrad:
Let's keep this concept, but revert the header, and move this to the footer, but only on screen sizes > lg

PR #371, opened 2025-11-26
Fix/collapse operation form on save

### LINKED ISSUE #353: autocollapse on save
Image

### LINKED ISSUE #366: Autocollapse on save
Need to autocollapse item in BoM and BoP in job, quote, and item master on save.

See this screen recording [here](https://drive.google.com/file/d/1Z3dUDP_OBPvG6ARwbm4Z_ZsE8WT1n00O/view?usp=drive_link).  When clicking "save," it should do the same thing it does as when i click "x" at the very end of the screen recording video.

jeebeez:
## What does this PR do?

This PR adds changes to autocollapse and show toasts when a user makes changes to a bill of operation or material.

- Fixes #366 
- Fixes #353 

https://github.com/user-attachments/assets/bc9290b4-4976-4379-b4f5-bce19aebb555

## How should this be tested?

- Try updating Bill of processes or materials anywhere in the product
- Saving them should autocollapse the form and show a toast

### DIFF
--- a/apps/erp/app/modules/items/ui/Item/BillOfMaterial.tsx
+++ b/apps/erp/app/modules/items/ui/Item/BillOfMaterial.tsx
@@ -36,7 +36,7 @@ import {
 import { AnimatePresence, LayoutGroup, motion } from "framer-motion";
 import { nanoid } from "nanoid";
 import type { Dispatch, SetStateAction } from "react";
-import { useCallback, useEffect, useState } from "react";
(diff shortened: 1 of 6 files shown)


sidwebworks (approved):
LGTM, I think later we should create something to wrap around useFetch maybe to handle this kind of repetitive logic where we are just checking on the state and doing a side-effect.

PR #428, opened 2026-01-10
Search Refactor

barbinbrad:
This PR refactors search to use a single table per customer. It uses the service role to bypass RLS. It also refactors the UI. The plan is in `llm/recommendations/global-search-architecture.md`

Screenshot 2026-01-10 at 2 56 12 PM


sidwebworks (approved):
Looks good and is working - just built the types and pushed one commit.

PR #432, opened 2026-01-13
feat: add tag filtering to inventory quantities page

### LINKED ISSUE #425: Allow filtering inventory data by item tags
### Is your proposal related to a problem?

 It is hard to filter items by tags on the inventory quantities page.

### Describe the solution you'd like

Allow filtering data by item tags in the inventory quantities table  

### Describe alternatives you've considered

N/A

### Additional context

Filtering by tag on items page - 
Image

Can't currently filter by tag on inventory quantities page - 
Image

### Requirement/Document

N/A

---

##### House rules

- If this issue has a `🚨 needs approval` label, don't start coding yet. Wait until a core member approves feature request by removing this label, then you can start coding.
  - For clarity: Non-core member issues automatically get the `🚨 needs approval` label.
  - Your feature ideas are invaluable to us! However, they undergo review to ensure alignment with the product's direction.
  - Follow Best Practices lined out in our [Contributor Docs](https://github.com/crbnos/carbon/blob/main/.github/CONTRIBUTING.md)

udayjain26:
## Summary

Add ability to filter inventory quantities by item tags, with support for distinguishing tags across different item types (Material, Part, Tool, Consumable).

**Changes:**
- Add `tags` column to `get_inventory_quantities` SQL function
- Add tags filter to InventoryTable with item type disambiguation
- Update `getTagsList` to return tag table/source information  
- Handle composite tag filters to enable filtering by both tag and type

**Key feature:** When the same tag name exists on different item types (e.g., "Press" on both Material and Consumable), the filter shows the source type (e.g., "Press (Material)") and filters correctly by both tag name and item type.

- [x]  I have self-reviewed the code (A decent size PR without self-review might be rejected).
- [x]  I confirm automated tests are in place that prove my fix is effective or that my feature works.

## Test Plan

- [x] Verified build passes (`npm run build`)
- [x] Tested locally: created tags on different item types, confirmed filter shows type disambiguation
- [x] Tested that selecting a specific tag (e.g., "Press (Material)") only filters items of that type with that tag

## Video

https://github.com/user-attachments/assets/6cd295c4-ae3a-4f06-87a6-ff15c43a5746

Fixes #425

### DIFF
--- a/apps/erp/app/modules/inventory/inventory.service.ts
+++ b/apps/erp/app/modules/inventory/inventory.service.ts
@@ -185,50 +185,7 @@ export async function getInventoryItems(
     );
   }
 
-  // Process tag filters with composite values (e.g., "Press|material")
-  // and expand them into separate tag and type filters
-  const processedArgs = { ...args };
-  if (args.filters) {
-    const typeFiltersToAdd: string[] = [];
-    processedArgs.filters = args.filters.map((filter) => {
-      if (filter.column === "tags" && filter.value) {
-        // Parse composite values: "tagName|table" format
-        const tagValues = filter.value.split(",");
-        const processedTags: string[] = [];
-
-        for (const tagValue of tagValues) {
-          if (tagValue.includes("|")) {
-            const [tagName, table] = tagValue.split("|");
-            processedTags.push(tagName);
-            // Capitalize first letter for type filter (e.g., "material" -> "Material")
-            const itemType = table.charAt(0).toUpperCase() + table.slice(1);
-            if (!typeFiltersToAdd.includes(itemType)) {
-              typeFiltersToAdd.push(itemType);
-            }
-          } else {
-            processedTags.push(tagValue);
-          }
-        }
-
-        return {
-          ...filter,
-          value: processedTags.join(",")
-        };
-      }
-      return filter;
-    });
-
-    // Add type filters if extracted from composite tag values
-    if (typeFiltersToAdd.length > 0) {
-      processedArgs.filters.push({
-        column: "type",
-        operator: "in",
-        value: typeFiltersToAdd.join(",")
-      });
-    }
-  }
-
-  query = setGenericQueryFilters(query, processedArgs, [
+  query = setGenericQueryFilters(query, args, [
     { column: "readableIdWithRevision", ascending: true }
   ]);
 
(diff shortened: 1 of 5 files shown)


crudmachine:
@udayjain26 thank you for taking this on. Some feedback - 
* You could undo all the changes to packages/database/src/types.ts‎ file in this PR except this one line https://github.com/crbnos/carbon/pull/432/changes#diff-13c08efd70f2041dbe510215d47563f7e73e21ce571928d7605685ecbb98d1f9R53181
* I thought about this more, since there is already an `Item Type` filter on the Quantities page, we should just keep the tags list as just the tag name to simplify things. If there are multiple tags with same name from different item types then we could just display 1 tag name for that (basically convert the `list` of the tag names across all item types to a `Set`). This will simplify the PR quite a bit. 
@barbinbrad what do you think about this point^? Should we keep the tags filter here like the PR currently has it like `{tag_name} ({item_type})` or should we instead just display the list of all the tags like `{tag_name}` ?

barbinbrad:
> @udayjain26 thank you for taking this on. Some feedback -
> 
> * You could undo all the changes to packages/database/src/types.ts‎ file in this PR except this one line https://github.com/crbnos/carbon/pull/432/changes#diff-13c08efd70f2041dbe510215d47563f7e73e21ce571928d7605685ecbb98d1f9R53181
> * I thought about this more, since there is already an `Item Type` filter on the Quantities page, we should just keep the tags list as just the tag name to simplify things. If there are multiple tags with same name from different item types then we could just display 1 tag name for that (basically convert the `list` of the tag names across all item types to a `Set`). This will simplify the PR quite a bit.
>   @barbinbrad what do you think about this point^? Should we keep the tags filter here like the PR currently has it like `{tag_name} ({item_type})` or should we instead just display the list of all the tags like `{tag_name}` ?

Let's jsut go with tag name so keep it consistent so we don't need patch code.

udayjain26:
@crudmachine @barbinbrad                                                                                                                                                                  

Thanks for the feedback! I've pushed a new commit that addresses your suggestions:                                                                                 

Changes made:                                                                                                                                                                                        Simplified the tags filter to show just {tag_name} (deduplicated using a Set)                                                                                                                      
Removed the composite tag value logic                                                                                                      
Simplified types.ts

Let me know if there's anything else you'd like me to adjust!

barbinbrad:
Looks good @udayjain26 , nice work 🫡

Issue #440, opened 2026-01-19
Implement Deactivating/Soft Deleting Processes (similar to Work Center Deactivation)


akshaylulla:
### Is your proposal related to a problem?

The list of Processes in the Processes dropdown in a Bill of Process can get pretty crowded (hard to search/find the process you want). 
Some Processes might not be used anymore at a business/company. We have a delete process functionality, but we don't allow deleting a process if it is used on a Bill of Process for an item. 

### Describe the solution you'd like

Similar to how we allow deactivating a work center, we should also give users the ability to deactivate processes. 

### Describe alternatives you've considered

I haven't considered other solutions.

### Additional context

(Write your answer here.)

### Requirement/Document

(Share it here.)

---

##### House rules

- If this issue has a `🚨 needs approval` label, don't start coding yet. Wait until a core member approves feature request by removing this label, then you can start coding.
  - For clarity: Non-core member issues automatically get the `🚨 needs approval` label.
  - Your feature ideas are invaluable to us! However, they undergo review to ensure alignment with the product's direction.
  - Follow Best Practices lined out in our [Contributor Docs](https://github.com/crbnos/carbon/blob/main/.github/CONTRIBUTING.md)

carbon-agent:
## 🛠️ Grooming Note

This feature appears to be **already implemented** in the codebase:

1. **Database:** Migration `20260120171236_process-active-column.sql` adds an `active` boolean column (default `true`) to the `process` table, and the `processes` view is recreated to include it.
2. **Service:** `processDeactivate()` in `resources.service.ts` sets `active: false`.
3. **Routes:** Both `processes.deactivate.$processId.tsx` and `processes.activate.$processId.tsx` exist with full loader/action/UI.
4. **UI:** `ProcessesTable` has activate/deactivate actions in the row context menu.

**Recommendation:** This issue can likely be closed as completed. If there are remaining gaps (e.g. filtering inactive processes from specific dropdowns), please reopen with the specific scope.

PR #445, opened 2026-01-20
feat: implement process deactivation (soft delete)

udayjain26:
## Summary
- Add ability to deactivate processes (soft delete) similar to work centers
- Deactivated processes are hidden from dropdowns but remain in the database for historical records
- Users can reactivate processes when needed

## Changes
- Add `active` column to process table with database migration
- Recreate processes view to include the new column
- Add `activateProcess` service function and update `deleteProcess` to do soft delete
- Filter `getProcessesList` to only return active processes (hides from dropdowns)
- Add Active column to ProcessesTable with filter capability
- Add Deactivate/Activate context menu options based on current state
- Create `processes.activate` route for reactivation

## Test plan
- [x] Go to Resources > Processes
- [x] Click on a process row menu and see "Deactivate Process" option
- [x] Deactivate a process and verify it shows as inactive
- [x] Filter by Active status to see inactive processes
- [x] Reactivate a process using "Activate Process" option
- [x] Verify deactivated processes don't appear in Bill of Process dropdowns

https://github.com/user-attachments/assets/ec0520cc-dff7-4424-9553-d769c1c42459

Builds #440

### DIFF
--- a/apps/erp/app/utils/path.ts
+++ b/apps/erp/app/utils/path.ts
@@ -1581,7 +1581,9 @@ export const path = {
     workCenterActivate: (id: string) =>
       generatePath(`${x}/resources/work-centers/activate/${id}`),
     activateProcess: (id: string) =>
-      generatePath(`${x}/resources/processes/activate/${id}`)
+      generatePath(`${x}/resources/processes/activate/${id}`),
+    deactivateProcess: (id: string) =>
+      generatePath(`${x}/resources/processes/deactivate/${id}`)
   }
 } as const;
 
--- a/apps/erp/app/modules/resources/resources.service.ts
+++ b/apps/erp/app/modules/resources/resources.service.ts
@@ -135,11 +135,18 @@ export async function activateProcess(
   return client.from("process").update({ active: true }).eq("id", processId);
 }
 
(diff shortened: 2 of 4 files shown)


apps/erp/app/utils/path.ts
```diff
@@ -1579,7 +1579,11 @@ export const path = {
     workCenter: (id: string) =>
       generatePath(`${x}/resources/work-centers/${id}`),
     workCenterActivate: (id: string) =>
-      generatePath(`${x}/resources/work-centers/activate/${id}`)
+      generatePath(`${x}/resources/work-centers/activate/${id}`),
+    activateProcess: (id: string) =>
+      generatePath(`${x}/resources/processes/activate/${id}`),
+    deactivateProcess: (id: string) =>
+      generatePath(`${x}/resources/processes/deactivate/${id}`)
```
barbinbrad: These should be alphabetized. Maybe rename them `processActivate` and processDeactivate` so that we can group them together

barbinbrad:
I was unable to push updaets to your branch so I made the branch within carbon org and merged. Closing this one. Thanks Uday!

PR #491, opened 2026-02-10
[FEAT] Add item stock quantities

jeebeez:
## What does this PR do?

- Added item stock quantities table for showing inventory
- Added cron to update it every half hour
- Updated realtime db to update it every 10 min

Screenshot 2026-02-10 at 3 06 59 PM

## How should this be tested?

- Load this PR locallly
- Updated inventory for an item from item master
- Ideally also post an invoice or do a negative adjustment to the item ledger
- Run the trigger job from Trigger(unless you want to wait 30 min, go brew something)
- When adding a line item you should see the stock quantity with the description

## Checklist


packages/database/supabase/migrations/20260210052714_item-stock-quantities-materialized-view.sql
```diff
@@ -0,0 +1,24 @@
+CREATE MATERIALIZED VIEW "itemStockQuantities" AS
+SELECT
+  "itemId",
+  "companyId",
+  SUM("quantity") AS "quantityOnHand"
+FROM "itemLedger"
+GROUP BY "itemId", "companyId";
```
barbinbrad: When you have multiple locations, it's not very useful to know the total quantity across all locations -- it's actually better to _not_ have a number in that case most of the time.

Can we group by locationId too and on the FE store the total quantity, and the quantity for each location? We should support passing an optional `locationId` to the `<Item>` component so that if we know the location (like in the case of a job/quote bill of material, a quote line, a sales order or purchase order line, etc) we show the quantity available for the location.

This adds a bit of scope I think. But the end result is much better.

For the sake of demoing this on Wednesday I'd be happy with just adding the location to the query, and showing the total in the Item so we didn't have to implement `locationId` prop on the Item in this ticket, but we could add that as a fast follow tomorrow.

PR #601, opened 2026-03-26
ci: split biome and types workflows; add turbo types task

sidgaikwad:
## Summary
This PR sets up dedicated quality workflows and fixes the missing Turbo 	ypes task so CI can run lint/format and type checks in a structured way.

## What Changed
- Added Turbo 	ypes task in 	urbo.json to match 
pm run types.
- Added root script ormat:check in package.json (iome format --check .).
- Added dedicated Biome workflow: .github/workflows/biome.yml.
  - Runs lint + format checks on PRs/pushes.
- Added dedicated Types workflow: .github/workflows/types.yml.
  - Runs 
pm run types in a separate workflow.
  - Marked as non-blocking for now (continue-on-error: true) to unblock current type debt cleanup in a follow-up PR.

## Why
- Keep quality gates split and easier to reason about (Biome vs Types).
- Improve signal in GitHub Actions by separating responsibilities.
- Preserve momentum while type issues are fixed incrementally.

## Follow-up (next PR)
- Resolve current type failures in @carbon/documents and cross-package type surface.
- Remove continue-on-error from 	ypes.yml once type checks pass consistently.
- Evaluate bundler migration (	sup -> 	sdown) as a dedicated change with package build/export alignment.

### DIFF
--- a/.github/workflows/biome.yml
+++ b/.github/workflows/biome.yml
@@ -1,0 +1,36 @@
+name: Biome
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: biome-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  lint-and-format:
+    name: Lint and Format
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Use Node.js 20.x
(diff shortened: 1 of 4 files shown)


barbinbrad:
Should we make merging dependent on the format and lint job passing?

sidgaikwad:
@barbinbrad  Yes, we should make lint/format required. This PR exposed existing repo-wide Biome issues, so I suggest we keep it non-blocking briefly (or scope to changed files), clean the baseline, then enforce it as required.

PR #668, opened 2026-04-15
Pricing Rules

jeebeez:
## What does this PR do?

- Added Pricing Rules to Carbon
- Added Pricing Lists to Carbon
- Add UI to show computed price traces in orders

https://github.com/user-attachments/assets/27225991-1232-4b6e-aaaa-701800b20189

## How should this be tested?

- Add rules for a particular seller and item
- Add price overrides for them
- Try creating a sales order or a quote using that item


barbinbrad:
I can't save a pricing rule, and the form element for adding an item doesn't work. Sending you a screen recording in Slack.

barbinbrad:
In pricing list, we should require a customer or a customer type through a zod refine. Currently you can create a rule that applies to nothing.

apps/erp/app/modules/sales/pricing/pricing.models.ts
```diff
@@ -0,0 +1,60 @@
+import { z } from "zod";
```
barbinbrad: don't add these .models.ts and .services.ts -- adjust the claude rule if neccessary, these should be in `sales.models` and `sales.service`

barbinbrad:
I don't get what we're doing here? There's no ability to add or update prices. They just show up there for every part and  render a price rule? What if I want to set the part to 2 for customer A, and 3 for customer B and 4 for customer C?

Screenshot 2026-04-15 at 8 22 21 PM

jeebeez:
For the pricelists I thought you asked for using the filters to select a customer / customer type to apply the rules instead of the action button towards the end but maybe I just got it wrong my bad.

apps/erp/app/components/Form/ChoiceCard.tsx
```diff
@@ -0,0 +1,112 @@
+import { cn, RadioGroup, RadioGroupItem } from "@carbon/react";
```
barbinbrad: Can we move this component to `@carbon/react` package?

apps/erp/app/modules/sales/ui/Customer/CustomerSidebar/useCustomerSidebar.tsx
```diff
@@ -70,6 +71,12 @@ export function useCustomerSidebar({ contacts, locations }: Props) {
       role: ["employee"],
       icon: <LuShieldAlert />
     },
+    {
+      name: t`Price List`,
+      to: `${path.to.salesPriceList}?customerId=${customerId}`,
```
barbinbrad: Can we remove this?

barbinbrad:
I'm not sure how you're deciding what rules to apply. In this case, it seems to me that both of these rules should have been applied in both cases? Or if we were choosing the "more specific one" -- I would think the other one would be applied.

Going to DM you the screen recording. It's too big

barbinbrad:
Should the prices be set automatically on a quote or just calculated?

Screenshot 2026-04-17 at 9 51 53 PM

barbinbrad (approved):

PR #683, opened 2026-04-21
Price List Refactor

jeebeez:
## What does this PR do?

- Updated how price lists work
- added duplication
- some changes to audit logs, etc

## How should this be tested?

- You should try adding items to a customers price list
- Try duplicating price lists' and individual items within it 
- Rest of the price list functionality and price break logs

## Checklist


apps/erp/app/modules/sales/sales.models.ts
```diff
@@ -181,6 +181,15 @@ export const priceOverrideValidator = z
     path: ["validTo"]
   });
 
+export const duplicatePriceListValidator = z.object({
```
sidwebworks: Why do we define this but not use it?

apps/erp/app/modules/sales/sales.service.ts
```diff
@@ -2227,30 +2232,45 @@ export async function resolvePriceList(
...
 
+  let scopeQuery = client
+    .from("customerItemPriceOverride")
+    .select("itemId")
+    .eq("companyId", companyId)
+    .eq("active", true);
+
+  if (args.customerId) {
+    scopeQuery = scopeQuery.eq("customerId", args.customerId);
+  } else if (args.customerTypeId) {
```
sidwebworks, line 2249: What if it's neither?

apps/erp/app/modules/sales/sales.service.ts
```diff
@@ -2822,6 +2824,126 @@ export async function deleteCustomerItemPriceOverride(
...
+    : target.customerTypeId
+      ? existingLookup.eq("customerTypeId", target.customerTypeId)
+      : existingLookup.is("customerId", null).is("customerTypeId", null);
+
+  const { data: existingOverrides } = await existingLookup;
+  const existingByItemId = new Map(
+    (existingOverrides ?? []).map((e) => [e.itemId, e.id])
+  );
+
+  for (const src of sourceOverrides) {
```
sidwebworks: we should begin a transaction here

apps/erp/app/routes/x+/sales+/price-list.duplicate.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const targetCustomerTypeId =
+    (formData.get("targetCustomerTypeId") as string) || undefined;
+  const conflictStrategy =
+    (formData.get("conflictStrategy") as "skip" | "overwrite") || "skip";
+
+  let overrideIds: string[] | undefined;
+  const overrideIdsRaw = formData.get("overrideIds") as string;
+  if (overrideIdsRaw) {
+    try {
+      overrideIds = JSON.parse(overrideIdsRaw);
```
sidwebworks: We should use a validator here I think

Also maybe do
```
overrideIds: overrideIds?.length ? overrideIds : undefined
```

sidwebworks (dismissed):
Looks good just a few comments

apps/erp/app/modules/sales/sales.service.ts
barbinbrad: agree

barbinbrad: do we need to put transactions in a `sales.server.ts`?

apps/erp/app/modules/sales/sales.service.ts
jeebeez: Shouldn't happen but yeah better to add an early return here

apps/erp/app/modules/sales/sales.models.ts
jeebeez: It was being used on the FE, added it to the action as well

apps/erp/app/modules/sales/sales.service.ts
jeebeez: Yup, a transactions makes sense here. Also yeah lets put put it in a `.server.ts`

sidwebworks (approved):

PR #692, opened 2026-04-22
Shelf Life

sidwebworks:
## What does this PR do?

Adds end-to-end shelf-life management for items produced via manufacturing jobs, plus a hierarchical Storage Units model and an Expired-entity consumption policy.

### Per-item shelf-life config

Items (Parts, Materials, Tools, Consumables) gain a Shelf Life section with four modes:

- **Not managed** — no expiry tracking.
- **Fixed Duration** — `today + N days` stamped on output. Optionally scope stamping to a specific trigger process. New flag *Inherit earliest input expiry* caps the result by the earliest input's expiry (food / pharma — flour can't outlive its rice).
- **Calculated** — inherits MIN expiry across consumed inputs. Company-level scope knob (`AllInputs` vs `ManagedInputsOnly`) decides whether Set-on-Receipt inputs feed MIN.
- **Set on Receipt** — supplier-stamped expiry captured at goods-receipt time on receipt lines (raw materials).

### Tracked-entity expiration column

Promoted `expirationDate` from a JSONB attribute to a first-class `DATE` column on `trackedEntity`, with a partial index for near-expiry reports. Edit-expiry modal records every override (previous, next, reason, user, timestamp) into `attributes.expiryOverrides` for audit.

### Expiry stamping & traceability

- DB function `set_shelf_life_for_operation` fires when an operation is marked Done and writes the computed expiry to the seed entity. Idempotent — replays don't shift an already-set date.
- New `ExpiryTracePopover` shows the policy chain that produced a given expiry (source date, mode, days, trigger process, input MIN, manual overrides).
- Tracked Entities table gains an Expiry column. **Amber** badge for rows within `nearExpiryWarningDays`, **red** badge for past-expiry rows.

### Expired-entity consumption policy

New company setting `expiredEntityPolicy` (default **Block**) controls what happens when an operator tries to consume a tracked entity past its expiry:

- **Warn** — record a warning, allow consumption.
- **Block** — reject the issue / transfer.
- **BlockWithOverride** — require an override reason, log it.

Wired into both edge functions (`issue`, `post-stock-transfer`) and the MES Issue Material modal (expired serials/batches show an *Expired* badge and are dropped from the picker under Block).

### Hierarchical Storage Units

- Storage Units are now nested via `parentId`, with depth + ancestor-path tracking.
- Items gain a *Default Storage Unit* (any depth in the hierarchy; `locationId` derived server-side).
- Storage Types associate to units via a tags-style multi-select.
- Receipt Lines and PickMethod forms can create new storage units inline.

### Company settings consolidation

All shelf-life knobs collapsed into `companySettings.inventoryShelfLife` JSONB:

```
{
  "calculatedInputScope":  "AllInputs" | "ManagedInputsOnly",
  "expiredEntityPolicy":   "Warn" | "Block" | "BlockWithOverride",
  "nearExpiryWarningDays": number | null,
  "defaultShelfLifeDays":  number
}
```

Flat columns dropped in the same migration. Settings page has a single Inventory section with all four knobs.

## Database migrations

7 branch-only migrations after consolidation:

- `20260420000000_item-storage-defaults-shelf-life` — `itemShelfLife` table, default storage unit on `pickMethod`, Storage Unit hierarchy + Storage Types, company-level shelf-life columns.
- `20260426000000_tracked-entity-item-fk` — direct FK from `trackedEntity` to `item`.
- `20260426020000_tracked-entity-expiration-date-column` — promotes `expirationDate` to a DATE column + partial index, redefines receipt-line / shelf-life functions to write the column directly.
- `20260426040000_calculated-shelf-life-input-scope` — adds `inventoryShelfLife` JSONB blob, branches `set_shelf_life_for_operation` on `calculatedInputScope`.
- `20260427010000_inventory-shelf-life-settings-and-inherit-input-expiry` — backfills JSONB from flat columns, drops them, adds `inheritEarliestInputExpiry` + CHECK constraint, redefines stamp helper to honor the flag.
- `20260427130000_shelf-life-start-on-receipt` — `Set on Receipt` mode for raw materials at goods-receipt time.
- `20260428100000_propagate-item-readable-id-to-tracked-entity` — keeps `trackedEntity.sourceDocumentReadableId` in sync with `item.readableIdWithRevision`.

## How should this be tested?

1. On any Part/Material/Tool/Consumable → **Inventory** tab → set Shelf-life mode to **Fixed Duration**, enter days (e.g. 7), optionally pick a trigger process. Save.
2. Create a job for that item with a make method containing the trigger operation.
3. Mark the operation **Done** → Inventory → Tracked Entities → output entity's Expiry column should show `today + N days`.
4. Settings → Inventory → set Near-expiry warning days larger than N → reload table → row shows **amber** badge.
5. Set `expirationDate` to the past on a row → row shows **red Expired** badge.
6. Settings → Inventory → set Expired-entity policy to **Block**. In MES, try to issue an expired serial — should be rejected. Switch to **BlockWithOverride** and confirm the override flow records a reason.
7. *Inherit earliest input expiry*: enable on a Fixed Duration BoM-parent. Issue an input with a near-term expiry; complete the operation. Output expiry should be the earlier of `today + days` and the input expiry.
8. **Calculated** mode: complete a job whose only input is a Set-on-Receipt raw material with a supplier-stamped expiry. Confirm the output inherits that expiry under default scope (`AllInputs`).
9. **Storage Units**: nest units via *Parent*; create one inline from a Receipt Line picker; assign as Default Storage Unit on an item.
10. **Edit Expiry**: open the popover on a Tracked Entity row → pick a new date + reason → confirm the override is logged in `attributes.expiryOverrides`.

## Mandatory Tasks

- [ ] I confirm automated tests are in place that prove my fix is effective or that my feature works

## Checklist

- [x] No new TypeScript/lint warnings introduced
- [x] DB stamp helper is idempotent (re-marking Done doesn't overwrite an existing expiry date)
- [x] Setting `nearExpiryWarningDays` to null/blank disables badges company-wide
- [x] Expired-entity policy enforced in both `issue` and `post-stock-transfer` edge functions
- [x] All raw `Date` operations replaced with `@internationalized/date` for timezone correctness


apps/erp/app/modules/items/ui/Item/ItemStorageAndShelfLifeFields.tsx
```diff
@@ -0,0 +1,128 @@
...
+              }))}
+            placeholder="Not managed"
+          />
+          {shelfLifeMode === "ItemSpecific" && (
+            <>
+              <NumberControlled
+                name="shelfLifeDays"
+                label="Shelf-life (days)"
+                minValue={1}
+                value={shelfLifeDays ?? 7}
```
jeebeez: nit: should the 7 be a constant?

apps/erp/app/modules/production/ui/Jobs/JobMaterialsTable.tsx
```diff
@@ -55,433 +56,467 @@ import type { Job, JobMaterial } from "../../types";
...
+                  </Hyperlink>
+                  {nearExpiryWarningDays !== null &&
+                    nearExpiryWarningDays !== undefined &&
+                    row.original.hasExpiredBatch && (
+                      <Badge
+                        variant="destructive"
+                        className="gap-1 text-xs shrink-0"
+                      >
+                        <LuCalendarX className="size-3" />
+                        Expired batch
```
jeebeez: Need `<Trans>`  here. Might be other instances Claude should be able to find and replace

jeebeez:
Too much text imo. 

Screenshot 2026-04-22 at 9 59 03 PM

Something like this.

Screenshot 2026-04-22 at 10 01 50 PM

It can be complex but I think it might be confusing.

jeebeez:
Should we confirm if the shelf trigger process is on the BOP before saving it. Could be an issue if someone deletes the process but fails to update this. I think it fails silently now

packages/database/supabase/migrations/20260420000000_item-storage-defaults-shelf-life.sql
```diff
@@ -0,0 +1,435 @@
...
+CREATE POLICY "Employees with parts_update can update item shelf life" ON "itemShelfLife"
+  FOR UPDATE
+  USING (
+    has_role('employee', "companyId")
+    AND has_company_permission('parts_update', "companyId")
+  );
+
+CREATE POLICY "Employees with parts_delete can delete item shelf life" ON "itemShelfLife"
+  FOR DELETE
+  USING (
```
barbinbrad: Use the convention from `rls-refactor` migration to name the policies `SELECT`, `INSERT`, `UPDATE` and `DELETE` -- don't use these `has_role` or `has_company_permission` functions either they are inefficient.

packages/database/supabase/migrations/20260420000000_item-storage-defaults-shelf-life.sql
```diff
@@ -0,0 +1,435 @@
...
+    has_role('employee', "companyId")
+    AND has_company_permission('parts_delete', "companyId")
+  );
+
+CREATE POLICY "Requests with an API key can access item shelf life" ON "itemShelfLife"
+  FOR ALL USING (has_valid_api_key_for_company("companyId"));
```
barbinbrad: Remove this is handled by the new style policy functions.

apps/erp/app/modules/items/items.models.ts
```diff
@@ -52,6 +52,13 @@ export const itemReplenishmentSystems = [
   "Buy and Make"
 ] as const;
 
+export const shelfLifeModes = [
```
barbinbrad, line 55: Are these an enum? Should they be?

apps/erp/app/modules/items/ui/Item/ItemStorageAndShelfLifeFields.tsx
```diff
@@ -0,0 +1,128 @@
...
+    case "ItemSpecific":
+      return "Item specific";
+    case "Calculated":
+      return "Calculated from BoM";
+    case "SetAtReceipt":
+      return "Set at receipt";
+  }
+};
+
+const ItemStorageAndShelfLifeFields = () => {
```
barbinbrad: I think this component is too much information to ask for on the initial form. I can see why we want a default shelf, but asking all this nitty gritty details seems crazy to me. Shelf life is so rare. It should be handled on the inventory tab I think.

packages/database/supabase/migrations/20260420000000_item-storage-defaults-shelf-life.sql
```diff
@@ -0,0 +1,435 @@
...
+--      be null; expiry is inherited from sub-assembly component batches only
+--      (SetAtReceipt inputs are excluded from the computation).
+--    - SetAtReceipt: expiry is entered by the user at receipt time (supplier-
+--      stated date). days and triggerProcessId must both be null. The stamp
+--      interceptor is a no-op for this mode.
+-- ----------------------------------------------------------------------------
+CREATE TABLE "itemShelfLife" (
+  "itemId"            TEXT NOT NULL,
+  "mode"              TEXT NOT NULL
+    CHECK ("mode" IN ('ItemSpecific', 'Calculated', 'SetAtReceipt')),
```
barbinbrad: Make this an enum. Add spaces to follow the convention. We don't use PascalCase anywhere.

apps/erp/app/modules/items/ui/Item/PickMethodForm.tsx
```diff
@@ -101,6 +136,59 @@ const PickMethodForm = ({
...
+        name="shelfLifeMode"
+        label="Shelf-life management"
+        options={shelfLifeModes
+          .filter((mode) => mode !== "NotManaged")
+          .map((mode) => ({
+            label: shelfLifeLabel(mode),
+            value: mode
+          }))}
+        placeholder="Not managed"
+      />
```
barbinbrad: Can we improve this UI. This dropdown is super confusing to me. I don't know what Item Specific or Calculated from BoM mean. Item Specific seems to mean a few different things? Maybe we can start with a small boolean component to say whether the shelf has a shelf-life. And then if it is active, we can use Gaurav's choices component with `col-span-3` to explain what's going on. Is the business logic different if the part replenishment is Make vs. Buy? Can we limit the choices based on that?

barbinbrad (changes requested):
I think the business logic and data model is good, but the UI is extremely unintuitive from my perspective. I'm not sure how to approach it. Maybe it's a better form, maybe it's a wizard, but I think we've really got to hold people's hand through this, and give feedback for what they've got configured in plain English. Currently "Set on receipt" is the only thing that I understand, and that doesn't seem to work. 

Here's a what I see when I try to receive a serial part:

Image

There's nothing about expiration date, and no matter what serial number I enter it seems to be "used already" which looks like a bug

barbinbrad:
Can we add this field in the zero fork? This is inappropriate for 99% of cases.

Screenshot 2026-04-23 at 9 46 12 PM

barbinbrad:
We should use the DatePicker component here. And we should use a 3 col layout so that the input matches the first input of the serial number

Screenshot 2026-04-23 at 9 51 01 PM

barbinbrad:
The expiration date isn't saving either. I'll DM you the video.

barbinbrad:
Looks like the expiry date doesn't populate for fixed date parts either: 
Screenshot 2026-04-23 at 10 20 29 PM

apps/erp/app/modules/inventory/ui/Traceability/TrackedEntitiesTable.tsx
```diff
@@ -109,6 +111,44 @@ const TrackedEntitiesTable = memo(
...
+              row.original.attributes as Record<string, unknown> | null
+            )?.expirationDate as string | undefined;
+            if (!expiry || nearExpiryWarningDays === null) return null;
+            const today = new Date();
+            today.setHours(0, 0, 0, 0);
+            const expiryDate = new Date(expiry);
+            const daysLeft = Math.floor(
+              (expiryDate.getTime() - today.getTime()) / 86_400_000
+            );
+            if (daysLeft < 0) {
```
barbinbrad: Try to use the `@internationalized/date` package (e.g. `cont today = today(getLocalTimezone())`

barbinbrad (changes requested):
Some UI cleanup stuff and getting the dates to work on the receipt. Haven't got to manufactured parts or actual expiry yet.

barbinbrad:
I made a part with PRE process expiration date setting, and 1/ there was no tracked entity before it was completed -- this seems wrong (I thought we made it immediately and put as status = Reserved), so it couldn't assign the expiration date I'm guessing. Then when I completed it, I could seethe tracked entity in the system. It has an expiration date in the attributes JSON but doesn't show an expiration date in the table.


https://github.com/user-attachments/assets/2622bf75-9bfd-4d97-9512-752465372b77

barbinbrad:
I think putting `expirationDate` in the JSON for `attributes` instead of as it's own column is not a good idea. JSON is nice for duping display data, but things that we want to use regularly, sort by, do business logic on should be their own column. I made this mistake with `Serial Number` and `Batch Number` originally, but then moved it to a first-class `readableId` column. I think we should do the same with `expirationDate`

barbinbrad:
When we receive a part with a fixed date expiration, it's populating the JSONB column but not the first-class `expirationDate` column. Also the `itemId` column isn't getting populated in the `trackedEntity` table.

barbinbrad:
When I split the line on the receipt so that I could have two different expirations, only one of the expiration dates got set.

barbinbrad:
When I split the line on the receipt so that I could have two different expirations, only one of the expiration dates got set.

barbinbrad:
When I split the line on the receipt so that I could have two different expirations, only one of the expiration dates got set.

barbinbrad:
The expiration date doesn't load on the receipt:
Screenshot 2026-04-27 at 10 16 26 PM

barbinbrad (approved):

PR #699, opened 2026-04-23
feat(erp): enable inline due date editing on schedule job cards

sidgaikwad:
## Summary
Adds inline due-date editing for job cards in Production Schedule (Dates view).

## What Changed
- Enabled inline due-date editing directly on schedule Job Cards.
- Wired updates to the existing schedule update action.
- Kept existing permissions and rescheduling behavior.
- Added optimistic column hint handling for smooth pending-state movement.
- Added optional `closeOnSelect` support to shared DatePicker and enabled it for Job Card picker.

## Why
Allows planners to adjust due dates from the schedule board without opening each job, reducing context switching.

## Testing
- [x] Typecheck
- [x] Manual verification in `/x/schedule/dates`
  - Select date from Job Card and verify update
  - Verify picker closes on selection
  - Verify data persists after refresh


barbinbrad:
Looks great

PR #703, opened 2026-04-25
fix for the deletion of already existing account

sidgaikwad:
## Changes 
- removed the deactiveAccount, deactiveEmployee function when user already exists
- retruned a simple error stating user already exists in the database
- now the working environment doesn't get nuked for the email


apps/erp/app/modules/users/users.server.ts
```diff
@@ -371,6 +374,20 @@ export async function createEmployeeAccount(
...
+      .eq("id", userId)
+      .eq("companyId", companyId)
+      .maybeSingle();
+
+    if (existingEmployee.data) {
+      return {
+        success: false,
+        message: "This user is already an employee in this company"
+      };
+    }
```
barbinbrad, line 387: I get why you're doing this, but I don't get why we want any of the other stuff?

sidgaikwad: changes done

apps/erp/app/modules/users/users.server.ts
```diff
@@ -443,7 +455,11 @@ export async function createEmployeeAccount(
     if (isNewUser) {
       await deleteAuthAccount(serviceRole, userId);
     } else {
-      await deactivateEmployee(serviceRole, userId, companyId);
+      // Clean up the employee and job records that were successfully inserted
+      await Promise.all([
```
jeebeez: Let's not do this we do a lot more than just delete these. Also I think its a soft delete

jeebeez (approved):

barbinbrad (approved):

apps/erp/app/modules/users/users.server.ts
jeebeez: this one

sidgaikwad: done with the changes

PR #768, opened 2026-05-12
Fix mobile overflow in digital quote share page

sidgaikwad:
## Summary
- fix digital quote line-item layout so text no longer bleeds off the right edge on mobile
- improve shared quote header responsiveness for small screens
- switch shared quote thumbnails from embedded base64 data URIs to signed storage URLs to reduce renderer memory pressure

## Notes
- this commit also includes locale catalog updates generated by the repository commit hooks


barbinbrad (changes requested):
I'm getting a tab crash every time I open the digital quote: 

https://github.com/user-attachments/assets/fc24fcda-04bc-401a-aa15-feea8aa7e97f

PR #778, opened 2026-05-14
feat(traceability): improve readAbleId search disambiguation with job/tracking context

sidgaikwad:
## Summary
- adds VIN-aware traceability search support in backend query matching
- enriches traceability search results with job context (`jobReadableId`) for easier disambiguation
- updates traceability dropdown rows to show `Job` and `Tracking` details while preserving existing result content
- adds truncated tooltip behavior via `TruncatedTooltipText` so long job/tracking values are fully readable on hover
- extends node search dialog matching to include VIN attribute values

## Problem
When searching by VIN/item-readable-id, users can get multiple tracked entity results and cannot reliably distinguish which row to open. This forced users to search by raw tracked entity id.

## Solution
- keep current search behavior and fields
- add disambiguating metadata (`Job` + `Tracking`) in dropdown rows
- include VIN key variants in search matching (`VIN`, `Vin`, `vin`, `VIN Number`, `Vehicle Identification Number`)
- show full values on hover only when text is truncated

## Testing
- `npm run typecheck -w erp`
- manual validation on traceability page:
  - search by VIN/item-readable-id returns multiple entities
  - each row shows job and tracking context
  - hovering truncated lines reveals full tooltip
  - selecting different rows opens corresponding entity graph

## Video Recording

https://github.com/user-attachments/assets/ae46b3f5-9c31-42ed-8222-3b67b5b62da2


sidwebworks (changes requested):
Requested changes on slack

barbinbrad:
I don't think this is the right approach. It looks like you are using the job id attribute to get the job readable id. But the right approach imo is to use the `trackedEntity.readableId` column.

sidgaikwad:
> I don't think this is the right approach. It looks like you are using the job id attribute to get the job readable id. But the right approach imo is to use the `trackedEntity.readableId` column.

@barbinbrad  Fixed it. The search already included `trackedEntity.readableId` (traceability.search.ts (line 38).) also the example video in pr description used the readableid (partid), but I also had extra job enrichment step that read `attributes.Job`, queried `job.jobId`, and showed that in the dropdown/recent results. That job enrichment made it look like the readable ID was coming from the job attribute path. I removed the job enrichment so the dropdown context now uses the tracked entity `readableId` directly.

barbinbrad (approved):

PR #779, opened 2026-05-15
feat: crbn up --borrow, realtime retry + toast, hydrate fixes

yidongw:
crbn up:
  - Add --borrow flag: pick another worktree's running containers instead
    of booting a new stack; shows a picker when multiple worktrees exist
  - Borrow uses the target's backend ports (DB, API, Inngest) but keeps
    own PORT_ERP/PORT_MES so dev servers don't conflict
  - Add portless mode support: conditionally run ensurePortlessInstalled,
    setupPortless, and pass portless flag through to spawnApps/renderEnv

  useRealtimeChannel:
  - Add toast notifications for CHANNEL_ERROR / TIMED_OUT (12s debounce)
    and setup failures, with notifyOnSubscribeError option (default: true
    in dev, false in prod)
  - Add exponential backoff retry (5s → 60s) for CHANNEL_ERROR / TIMED_OUT;
    CLOSED now only tears down, no longer retries
  - Extract subscribe logic to doSubscribeRef so retry timers always call
    the latest closure; cancel timer on cleanup
  - Remove console.log calls

  RealtimeDataProvider:
  - Add carbon and accessToken to hydrate effect deps (was only companyId),
    so hydrate re-runs when auth becomes available after initial mount
  - Add effect to reset hydratedFromServer on logout (accessToken → falsy)
  - IDB writes now awaited in parallel via Promise.all
  - Move error handling to .catch() at the call site

  Auth:
  - Add DEV_BYPASS_EMAIL: when set and email matches, skip magic link
    and sign in immediately via signInWithBypassEmail (generates +
    verifies the OTP token server-side)
  - Fix magic link and invite redirect URLs: /callback suffix was missing
  - Move Ratelimit instantiation inside action/loader (was module-level,
    causing redis connection at import time)
  Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

## What does this PR do?

## How should this be tested?

- Are there environment variables that should be set?
- What are the minimal test data to have?
- What is expected (happy path) to have (input and output)?
- Any other important info that could help to test that PR

## Checklist


apps/starter/app/routes/_public+/login.tsx
```diff
@@ -54,15 +54,14 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
-const ratelimit = new Ratelimit({
-  redis,
-  limiter: Ratelimit.slidingWindow(RATE_LIMIT, "1 h"),
-  analytics: true
-});
-
 export async function action({ request }: ActionFunctionArgs) {
   assertIsPost(request);
   const ip = request.headers.get("x-forwarded-for") ?? "127.0.0.1";
+  const ratelimit = new Ratelimit({
```
barbinbrad, line 60: why are you doing this?

yidongw: image

image

barbinbrad: I mean why are you moving the rate limit instantiation inside the action

yidongw: > The root cause was that const ratelimit = new Ratelimit({ redis, ... }) was at module level (not inside action), so React Router's Vite plugin didn't strip it from the client
  bundle. @carbon/kv uses ioredis which is Node.js-only — when the browser tried to import it, the module crashed, preventing hydrateRoot from ever completing. Zero JS on the page
  was the result.


because I have Zero JS on this verify page. all the numbsers are not working as expected

barbinbrad: Got it! Good one!

barbinbrad (approved):

PR #782, opened 2026-05-15
Filter using type when fetching and updating items

jeebeez:
## What does this PR do?

- Filter using type when fetching and updating items

## How should this be tested?

- try adding a part and consumable using the sequence or with the same readable id
- update the id for one, it should not do it for the other item with the same readable id
- also you should not see it in the sidebar as a revision while they have the same id

### DIFF
--- a/packages/database/supabase/migrations/20260515120000_fix-item-details-revisions-filter-by-type.sql
+++ b/packages/database/supabase/migrations/20260515120000_fix-item-details-revisions-filter-by-type.sql
@@ -1,0 +1,493 @@
+-- Fix cross-type leakage in get_part_details / get_tool_details /
+-- get_material_details / get_consumable_details.
+--
+-- These RPCs power the detail page loaders for each item type. Their
+-- item_revisions CTE pulled every "item" row sharing the same
+-- (readableId, companyId) regardless of type. Since the item uniqueness
+-- constraint is (readableId, revision, companyId, type), a Part and a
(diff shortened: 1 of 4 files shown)


naveenkash (approved):
LGTM

packages/database/supabase/migrations/20260515120000_fix-item-details-revisions-filter-by-type.sql
```diff
@@ -0,0 +1,493 @@
...
+  RETURN QUERY
+  WITH item_revisions AS (
+    SELECT
+      json_agg(
+        json_build_object(
+          'id', i.id,
+          'revision', i."revision",
+          'methodType', i."defaultMethodType",
+          'type', i."type"
+        ) ORDER BY
```
naveenkash, line 305: any reason not doing ```i."createdAt" DESC```

packages/database/supabase/migrations/20260515120000_fix-item-details-revisions-filter-by-type.sql
```diff
@@ -0,0 +1,493 @@
...
```
naveenkash, line 436: same as above

packages/database/supabase/migrations/20260515120000_fix-item-details-revisions-filter-by-type.sql
jeebeez: ordering is mostly done by the readable id

PR #785, opened 2026-05-18
MES operation time entry history editing

sidgaikwad:
# MES Operation Time Entry History Editing

## Summary
- Add operation-scoped editable time-entry history for completed MES production events.
- Move the full history into a modal behind a compact bottom-bar summary so the operation screen does not overlap.
- Keep active entries visible but read-only, and guard against duplicate active rows in realtime state.
- Add second-level editing with client and server validation for valid ranges, no future timestamps, and no overlapping intervals.
- Recalculate progress from realtime production-event updates and return the live event list from `useOperation`.

## Demo Video

https://github.com/user-attachments/assets/2e7c08c6-9fda-4d3e-b1ca-b17f89a99b92

## Validation
- `pnpm.cmd --filter mes typecheck`
- `npx.cmd biome check --formatter-enabled=false apps/mes/app/components/JobOperation/JobOperation.tsx apps/mes/app/components/JobOperation/hooks/useOperation.tsx apps/mes/app/routes/x+/event.tsx apps/mes/app/services/models.ts apps/mes/app/services/operations.service.ts`
- `git diff --check`
- `pnpm.cmd run build:mes`
- Confirmed no test file was added per request


barbinbrad:
It's uncommon to let shop-floor employees edit their timecards manually. I'd like to understand the goal here a little more.

PR #794, opened 2026-05-19
Add auth middleware + Reduce MCP layers

Samyak2406:
## What does this PR do?

  - Global auth middleware (clientScopeMiddleware + authContextMiddleware) wired into ERP, MES, and Academy roots. Service files read identity (userId, companyId, client) from
  AuthContextHolder (ALS) instead of taking them positionally.
  - Service files renamed *.service.ts → *.service.server.ts so client bundles can't reach server-only code.
  - MCP runtime simplified: removed the in-memory registry / bootstrap / mcp-tools.generated.server.ts stack. mcpTool() is now a build-time-only marker, mcp-tools.json (auto-generated) is
  the discovery index, and direct-executor.ts uses a static-import map for dispatch.


barbinbrad:
What's the motivation for `.service.ts` -> `.service.server.ts`? The code is idempotent

barbinbrad:
> What's the motivation for `.service.ts` -> `.service.server.ts`? The code is idempotent

It also makes the PR unmanageably large to review.

Samyak2406:
> What's the motivation for `.service.ts` -> `.service.server.ts`? The code is idempotent

@barbinbrad I found the rename .service.ts → .service.server.ts necessary. The chain:

  1. Service functions stopped taking client as a parameter and now read identity from AuthContextHolder / getAuthClient().
    - account.service.server.ts:4-8 imports them from ~/services/mcp/index.server.
  2. That barrel re-exports from @carbon/auth/auth-context.server, whose first line is:
    - import { AsyncLocalStorage } from "node:async_hooks" (packages/auth/src/services/auth-context.server.ts:1)
  3. node:async_hooks is server-only. Any service file is now transitively a server-only module.
  4. [React Router's bundler gurantees .server.ts suffix](https://reactrouter.com/api/framework-conventions/server-modules) to exclude a module from the client bundle. Without the rename, if any service file is reachable from a client component, the client build pulls in node:async_hooks -> build breaks / server code leaks. 

So the rename is what enforces the new server-only dependency that the ALS refactor introduced. Let me know your thoughts on it.

Samyak2406:
> > What's the motivation for `.service.ts` -> `.service.server.ts`? The code is idempotent
> 
> It also makes the PR unmanageably large to review.

I agree the with the fact PR has grown too large & hence difficult to be reviewed, as it involves all the services getting wrapped around with custom mcp wrapper & removing the client and auth params.

PR #798, opened 2026-05-20
Feat/add foreign key snapshots to audit logs

jeebeez:
## What does this PR do?

- Added Foreign key snapshots to audit logs
- Add prettification to label fields (can remove if we want just looked better)

How single field snapshots look

Screenshot 2026-05-20 at 12 30 12 PM

How multi-field snapshots will look (added just to demo shelf life doesn't need multi field snapshots)

Screenshot 2026-05-20 at 12 32 20 PM

## How should this be tested?

- Create item(Make, Batch tracking), add a couple processes to BOP
- Make sure audit logs are on 
- Turn on the shelf life from inventory and save
- Switch the process and save a couple times
- wait a mintue and check the logs

## Checklist


sidwebworks (approved):
Thinking if we should switch from createFields to this new snapshot fields? I like it better plus not many people use audit logs

PR #804, opened 2026-05-21
feat: re-link job to sales order from job properties

sidgaikwad:
## Summary

Companion to #800 (unlink confirmation modal). Even with that protection in place, jobs that were already unlinked before the fix — or edge cases in the future — have no way to restore the link through the UI. This PR adds that capability.

### `update.tsx`
Before clearing `salesOrderId` + `salesOrderLineId`, the previous values are now stored in `job.customFields` as `previousSalesOrderId`, `previousSalesOrderLineId`, and `previousSalesOrderReadableId`. This gives the relink modal something to show as "Previously linked".

### `$jobId.link.tsx` (new)
New action route at `/x/job/:jobId/link`. Accepts a `salesOrderLineId`, looks up the corresponding `salesOrderId` and `customerId` in a single query, then atomically updates all three fields on the job. Clears the history keys from `customFields` after a successful relink.

### `path.ts`
Added `jobLink(id)` entry.

### `JobProperties.tsx`
In the unlinked Target branch (where `salesOrderId` is null), a **Link to Sales Order** button now appears above the StorageUnit picker. Clicking it opens a modal with two cascading selects:

- **Sales Order** — open orders (`To Ship and Invoice` / `To Ship`) filtered to the job's customer. If history exists in `customFields`, the previously linked SO appears at the top under "Previously linked".
- **Line** — Make-to-Order lines on the selected SO filtered to the job's item (`itemId`). Label shows `PART-001 — qty 5 — Jun 15` so multiple lines are distinguishable.

The Link button stays disabled until a line is selected. Locked jobs (Completed / Closed / Cancelled) see the button disabled via the existing `isDisabled` flag.

## Test plan

- [ ] Open a job with `salesOrderId = null` (unlink one using the existing Unlink button).
- [ ] Target section shows **Link to Sales Order** button above StorageUnit picker.
- [ ] Click it → modal opens. Sales Order dropdown shows only open SOs for the job's customer. Previously linked SO appears at top if applicable.
- [ ] Pick an SO → Line dropdown populates with Make-to-Order lines matching the job's item.
- [ ] Pick a line → **Link** button enables.
- [ ] Click Link → Target field switches to the SO badge. DB: `salesOrderId`, `salesOrderLineId`, `customerId` all set. `customFields` history keys cleared.
- [ ] Navigate to the SO → job reappears in the "1 JOBS" badge on the line.
- [ ] On a locked job → Link button is disabled.
- [ ] Unlink again → check `customFields` has `previousSalesOrderId` stored. Reopen Link modal → previously linked SO shows at top.

## Notes

- Depends on #800 being merged first (both modify `JobProperties.tsx`).
- The `customFields` history keys are internal implementation detail — they are cleaned up after a successful relink so they don't pollute user-visible custom fields.


sidwebworks:
## Review — relink job to sales order

A few things to address before merge.

### Relink button never shows for the jobs this PR targets

In `JobProperties.tsx` the **Link to Sales Order** button is gated on `{!!prevSalesOrderId && (...)}`. Jobs that were unlinked before this fix have no `previousSalesOrderId` stored, so the button never renders for exactly the population it's meant to serve.

Gate the button on the unlinked condition (`salesOrderId == null` — the Target branch you're already inside), and let `previousSalesOrderId` drive only the "Previously linked" optgroup.

### Silent failure and lost atomicity in `update.tsx`

The unlink path was changed from a single `.update().in(ids)` to a select + N `Promise.all(updates)`, followed by an unconditional `return { data: null, error: null }`.

Supabase update errors resolve rather than reject, so `Promise.all` never throws — every failure is reported to the caller as success. Collect the results and return the first `.error` instead of hardcoding success. Also note the operation is no longer atomic across jobs, so a partial unlink is possible; keep it a single statement or wrap it in a transaction if that matters.

### Action trusts client input in `$jobId.link.tsx`

The action accepts `salesOrderLineId` and links it without verifying the line's `itemId` matches `job.itemId` or that `methodType = 'Make to Order'`. The modal filters on these, but the server doesn't, so a crafted request can link a job to an arbitrary line/item. Add those checks to the `salesOrderLine` lookup.

### Smaller items

- Use the design-system `Combobox`/`Select` instead of the raw `<select className="border border-input ...">` (x2). Per CLAUDE.md, prefer existing components — also gives search for customers with many open SOs.
- `getCarbonServiceRole()` is called inside the shipment loop in `$jobId.link.tsx` — hoist it above the loop.
- The shipment `functions.invoke("create", ...)` result is ignored, and `relinkedShipment = true` is set even when the invoke fails, so the "shipment lines restored" message can be wrong. Check the invoke result before setting the flag.
- The line label renders raw `line.promisedDate` — unlocalized date string. Format it like the other date displays.
- `$jobId.status.tsx` uses `(client as any)` on the query. If it's only to satisfy the regenerated view types, type the result instead (already done below) and drop the `as any`.
- A previously-linked SO can appear twice (in both optgroups) since they share the same `value` — filter `prevSalesOrderId` out of the open list.
- The PR description says history is stored in `job.customFields`, but the code adds real columns (`previousSalesOrderId` / `previousSalesOrderLineId` / `previousSalesOrderReadableId`) plus a migration. Update the description.

Top priority is the button gate — without it the feature does nothing for its target jobs.

sidwebworks (changes requested):
Requested changes

PR #815, opened 2026-05-25
Feat/improve bulk imports

jeebeez:
## What does this PR do?

- Improve bulk imports
   - Add new columns
   - Add automatching for fields
   - Add templates with instructions
- Add readable IDs to suppliers

## How should this be tested?

- Try downloading bulk import CSVs for Parts/Customers/Suppliers and import them
- The best order is Suppliers -> Customers(Optional) -> Supplier Contact and Parts


barbinbrad (changes requested):
Couple problems with supplier and customer IDs. 

1. It's annoying - we need a setting if we want to use it. I don't want other people being subjected to this.
2. It's editable like as a text box. It shouldn't be so easy to edit.

packages/database/supabase/migrations/20260521124731_supplier-customer-readable-id.sql
```diff
@@ -0,0 +1,232 @@
+-- Add a first-class human-readable identifier to supplier and customer.
```
barbinbrad, line 1: these migrations that end in 0000 are big problems, because if anyone else merges a migration from the same day supabase chokes because the timestamp is the unique key.

barbinbrad:
we need better testing notes on these. what fields are you adding?

barbinbrad:
> we need better testing notes on these. what fields are you adding?

Screenshot 2026-05-31 at 9 39 30 PM

although the `/pr-explainer` skill does a pretty good job

barbinbrad (approved):

PR #825, opened 2026-05-26
feat: add custom markup % input to quote line pricing dropdown

sidgaikwad:
## Summary

- Adds a custom text input at the bottom of the **Markup %** dropdown in the Quote Line Pricing card
- Users can now type any specific markup percentage (e.g. `23.5`) instead of being limited to the preset values (0%, 10%, 15%, 20%, etc.)
- Pressing **Enter** or clicking **Apply** triggers the recalculation

## Details

The custom input is rendered inside the dropdown's content panel with `stopPropagation` on click and keydown events so the dropdown stays open while typing. On apply, the same `onRecalculate` handler used by the preset items is called, keeping behavior consistent.

## Images
image

## Test plan

- [ ] Open a quote line in Draft status
- [ ] Click the **Markup %** button
- [ ] Type a custom value (e.g. `23.5`) in the input at the bottom
- [ ] Press Enter — prices should recalculate at 23.5% markup
- [ ] Click Apply instead — same result
- [ ] Verify preset items (10%, 20%, etc.) still work as before

### DIFF
--- a/apps/erp/app/modules/sales/ui/Quotes/QuoteLinePricing.tsx
+++ b/apps/erp/app/modules/sales/ui/Quotes/QuoteLinePricing.tsx
@@ -613,9 +613,14 @@ const QuoteLinePricing = ({
                     onKeyDown={(e) => e.stopPropagation()}
                   >
                     <NumberField
-                      value={customMarkup === "" ? undefined : Number(customMarkup)}
+                      value={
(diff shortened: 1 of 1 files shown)


barbinbrad:
Can you include a screenshot?

sidgaikwad:
> Can you include a screenshot?

yeah adding an example

sidgaikwad:
@barbinbrad, added the image

PR #828, opened 2026-05-27
fix(documents): show correct time units in job traveller PDF

sidgaikwad:
## Problem

Operation times in the job traveller PDF were always displayed as minutes regardless of the unit selected on the operation. This happened because `formatDurationMinutes()` was called unconditionally for setup, labor, and machine times — treating every value as minutes and formatting it as a duration string.

For example, with **Labor: 2 Hours/Piece**, the traveller would show `2 min` instead of `2 hr/pc`.

## Root Cause

`JobTravelerPDF.tsx` ignored the `setupUnit`, `laborUnit`, and `machineUnit` fields entirely and passed the raw numeric value straight into `formatDurationMinutes()`, which always assumes the input is in minutes.

## Fix

- Removed the `formatDurationMinutes()` calls for operation times
- Added a `factorAbbreviations` map (exhaustive over all `factor` enum values) that converts the full unit label to a compact abbreviation (e.g. `"Hours/Piece"` → `"hr/pc"`, `"Total Hours"` → `"hr"`)
- Time is now displayed as `{value} {abbreviatedUnit}`, which is correct for all unit types including rates (hr/pc, min/pc, pcs/hr, etc.)

The abbreviation map is typed as `Record<Factor, string>` so TypeScript will flag any missing entries if new units are added to the enum.

## Before / After
Before: 
image

After:
image

## Files Changed

- `packages/documents/src/pdf/JobTravelerPDF.tsx`

### DIFF
--- a/packages/documents/src/pdf/JobTravelerPDF.tsx
+++ b/packages/documents/src/pdf/JobTravelerPDF.tsx
@@ -1,7 +1,7 @@
 import type { Database } from "@carbon/database";
 import { getMESUrl } from "@carbon/env";
 import type { JSONContent } from "@carbon/react";
-import { formatDate } from "@carbon/utils";
+import { formatDate, formatFactor } from "@carbon/utils";
 import { Image, StyleSheet, Text, View } from "@react-pdf/renderer";
 import { createTw } from "react-pdf-tailwind";
 import { generateQRCode } from "../qr/qr-code";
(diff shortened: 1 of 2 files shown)


jeebeez:
This is nice. Can you make it a util and add it in the util file and then use it

jeebeez (approved):

PR #844, opened 2026-06-01
feat: remove approval for untouched MRP orders

jeebeez:
## What does this PR do?

- Added functionality to auto approve MRP orders

## How should this be tested?

-  Add an approval rule for an amount
- Create two orders from Purchasing/Planning that should trigger this rule
- Change a line item's quantity/price before finalizing it, and don't do anything on the other order
- The one with the change should trigger the approval rule.


jeebeez:
Does not seem like the right flow. Talked to Cambridge guys.

PR #858, opened 2026-06-04
Fix tax logic on invoices

jeebeez:
## What does this PR do?

-  Fixes dead column on invoices

## How should this be tested?

- Create a Sales Invoice with multiple line items that have tax
- Previously the totalTax field on the salesInvoice would still show 0, now it should show the right amount


barbinbrad:
Can you walk me through how tax is working? I'd figure that it'd be computed/derived, not stored. Do we need to store it?

jeebeez:
It is not computed; this saves it, but we could add a generated column here. Currently, the tax column on the invoice never updates, and this is bad, especially for people using the API. This fix is so that we can show the correct value there

barbinbrad:
> It is not computed; this saves it, but we could add a generated column here. Currently, the tax column on the invoice never updates, and this is bad, especially for people using the API. This fix is so that we can show the correct value there

Let’s make it generated/computed in a view, and have the backend methods for getting sales invoices use the view so that the MCP client will get the computed totals too.

barbinbrad (approved):

PR #886, opened 2026-06-12
feat: enhance job materials procurement status handling

naveenkash:
## What does this PR do?

Adds a per-material procurement / order-status indicator to a job's materials, so planners can see at a glance whether each material is covered, short, or in flight — without leaving the job.

Core changes
- Order-status badge on every job material, shown in both the Materials table and the BoM tree (JobBoMExplorer). States: Needs ordering, In stock, Planned purchase order, Awaiting approval, Received X of Y, On order, Planned job / Job in progress, and Completed.
- Status computed by priority-based allocation (getJobOrderStatusMap + jobOrderStatus.ts): on-hand stock and incoming supply (open POs, planned/approval-pending POs, and supply jobs) are allocated across active jobs by priority, so a remaining shortfall on a lower-priority job isn't masked by supply already spoken for. A single source of truth (getJobOrderStatusCategory) drives both the badge and the table's new Status filter.
- Materials-table row menu → planning: the row dropdown now links to Purchasing Planning (Buy items) or Production Planning (Make items), routed by the item's replenishment system and opened in a new tab. This replaces the old stock-transfer "Action Items" session flow, which is removed (StockTransferSessionWidget, its route, store session block, and path helper).
- DB: new migration 20260612134529_job-quantity-on-hand-null-storage-unit.sql (handles null storage unit in the job quantity-on-hand RPC) + regenerated types.

## How should this be tested?

- Are there environment variables that should be set?
- What are the minimal test data to have?
- What is expected (happy path) to have (input and output)?
- Any other important info that could help to test that PR

## Checklist


barbinbrad:
Looks great. Couple nits.

1/ can we add a status filter to the job materials page?

Screenshot 2026-06-14 at 9 24 24 PM

2/ can we hide these indicators when the job is canceled or completed or the part has already been issued? for example if i've issued the part, and that uses up all the inventory, and then i go to look at the job it shouldn't tell me it needs ordered because we've already used the parts that we had on the job. i think we have to look estimates and actuals for that.

Screenshot 2026-06-14 at 9 24 35 PM

barbinbrad (changes requested):
Just need to figure out when to _not_ show these now, and add a little filtering on the job materials page.

barbinbrad:
The good part was that purchased parts seemed to work well -- even stacking dependencies so that one job would eat the inventory, and it would say we're fine, but a second job would say we need to order. But manufactured parts seems very incomplete.

1. When the job was a draft I didn't see any status icons. This seems wrong. 
2. When I added a planned job for a manufactured part that was required, the status was still a red blinking dot. This seems weird to me. On the other hand when I added a planned PO for a purchased part that was required, it changed from a red dot to a blue circle

Screenshot 2026-06-22 at 11 13 13 PM

3. When I release the job for manufactured dependency there was no status icon at all

Screenshot 2026-06-22 at 11 15 48 PM

4. When I received the manufactured dependency to inventory there was no status icon at all.
5. The red dot and blue circle aren't visually consistent. We can use the same icon that we use for waiting instead of the pulsing red dot to solve this problem.

barbinbrad:
the job status isn't taking planned manufactured jobs into account  (neither is the incoming column in the job materials)-- only released jobs -- so when you do MRP to order it -- nothing happens, it just says to order them again and again

we need to take planned production into account, similarly to how we take planned purchased orders into account

packages/react/src/PulsingDot.tsx
```diff
@@ -18,9 +18,9 @@ export function PulsingDot({
   return (
     <span className={cn("relative flex h-2 w-2", className)} {...props}>
       <span
-        className={`absolute h-full w-full animate-ping rounded-full border border-emerald-500 opacity-100 duration-1000`}
+        className={`absolute h-full w-full animate-ping rounded-full border border-amber-500 opacity-100 duration-1000`}
```
barbinbrad: This is used in many places. And green is the right color in many places. Create some variants for this please following the patterns with `cva`

barbinbrad (approved):
please add the color prop to `<PulsingDot />` and then feel free to merge

PR #909, opened 2026-06-18
Issue 448

### LINKED ISSUE #448: Can't filter Sales Order by In Progress Status
Found a bug? Please fill out the sections below. 👍

### Issue Summary

A summary of the issue. This needs to be a clear detailed-rich summary.

### Steps to Reproduce

1. (for example) Went to ...
2. Clicked on...
3. ...

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

### Actual Results

- Can't filter sales orders by 'In Progress' status

### Expected Results

- Can filter sales orders by 'In Progress' status

### Technical details

- Browser version, screen recording, console logs, network requests: You can make a recording with [Bird Eats Bug](https://birdeatsbug.com/).
- Node.js version
- Anything else that you think could be an issue.

### Evidence

- How was this tested? This is quite mandatory in terms of bugs. Providing evidence of your testing with screenshots or/and videos is an amazing way to prove the bug and a troubleshooting chance to find the solution.

Image

aashu0148:
What does this PR do?

Makes "In Progress" a real, filterable status on the sales orders list.  There was no row in the DB with status = 'In Progress' - it's a UI-only concept meaning "this order has a Make-to-Order line whose jobs haven't completed the requested quantity."

This adds a computed displayStatus column to the salesOrders view that returns 'In Progress' when manufacturing is in flight (and the order isn't Closed/Cancelled), otherwise the raw enum. The table sorts and filters on displayStatus; status is left untouched.

Why a new column instead of overwriting status: overwriting would silently break every other filter - e.g. .eq("status", "Confirmed") would drop genuinely-Confirmed orders that happen to have a job in flight (the view would report them as "In Progress"). Two columns = two purposes: status stays the source of truth for the underlying enum; displayStatus drives the badge and filter.

- Fixes #448

Visual Demo
https://github.com/user-attachments/assets/dc7fe4c2-27b1-462b-8de1-b4eac86b9517

Mandatory Tasks
- [x] I have self-reviewed the code.
- [ ] Automated tests — not added; covered by manual verification.

How should this be tested?
Try applying multiples filters on the sales orders table. 

Checklist
- [x] Read the contributing guide
- [x] Code follows project style
- [x] Commented where intent isn't obvious from names
- [x] No new warnings

### DIFF
--- a/apps/erp/app/modules/sales/sales.models.ts
+++ b/apps/erp/app/modules/sales/sales.models.ts
@@ -678,7 +678,7 @@ export const salesOrderLineType = [
 
 export const salesOrderStatusType = [
   "Draft",
-  // "In Progress",
+  "In Progress",
   "Needs Approval",
   // "Confirmed",
   "To Ship and Invoice",
--- a/apps/erp/app/modules/sales/ui/SalesOrder/SalesOrdersTable.tsx
+++ b/apps/erp/app/modules/sales/ui/SalesOrder/SalesOrdersTable.tsx
@@ -154,38 +154,9 @@ const SalesOrdersTable = memo(({ data, count }: SalesOrdersTableProps) => {
         }
       },
       {
-        accessorKey: "status",
+        accessorKey: "displayStatus",
         header: t`Status`,
-        cell: ({ row }) => {
-          const status =
-            row.getValue<(typeof salesOrderStatusType)[number]>("status");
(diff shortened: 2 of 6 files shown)


barbinbrad (approved):

barbinbrad:
Please give a more descriptive title.

PR #911, opened 2026-06-19
feat(items): Import BOM & BOP (Method Import) via CSV

sidgaikwad:
## Import BOM & BOP (Method Import) via CSV

Carbon can bulk-import flat entities (parts, materials, customers…), but a part arrives with **no
Make Method** — no **BOM** (what it's built from) and no **BOP** (the operations to build it). Teams
migrating a catalog had to hand-rebuild every recipe in the UI, one line at a time. This PR teaches
the existing CSV import machine to also build a part's recipe.

One **row-type-multiplexed engine** backs three entry points in the existing **Bulk Import** dropdown
on the **Parts** and **Tools** lists:

| Option | `table` | Use |
|--------|---------|-----|
| **Import BOM** | `bom` | BOM lines against parts that already exist |
| **Import Operations** | `operations` | Operations + steps/tools/parameters against existing parts |
| **Import Parts with Methods** | `partWithMethod` | Combined: create parts **and** their full BOM + BOP |

A `Row Type` column discriminates six row kinds — `PART / BOM / BOP / STEP / TOOL / PARAM` — and every
non-PART row names its parent part explicitly (`Parent ID` + `Parent Revision` + optional
`Make Method Version`). Each operation carries an `Op No` unique within its parent; `STEP`/`TOOL`/`PARAM`
rows reference their operation via `(Parent + Op No)`. Multi-level BOM is recursive — a sub-assembly is
just another part that owns its own rows — so no fragile positional `Level` column is needed.

### Entry points → one engine

```mermaid
flowchart LR
    subgraph UI["Bulk Import dropdown (Parts / Tools)"]
        A["Import BOM"]
        B["Import Operations"]
        C["Import Parts with Methods"]
    end
    A -->|"table = bom"| R["import.$tableId route<br/>(permission + Zod)"]
    B -->|"table = operations"| R
    C -->|"table = partWithMethod"| R
    R --> EF["import-csv edge function"]
    EF --> ENG["method-import.ts engine"]
    RULE["imports.models.ts<br/>(field mappings + Zod per type)"] -. defines .-> R
```

### Processing flow

The whole file is validated before anything is written; each part's recipe then commits
**all-or-nothing** in its own transaction, so one bad line never produces a half-built recipe.

```mermaid
flowchart TD
    P["Parse CSV + apply column / enum mappings"] --> CL["Classify each row by Row Type"]
    CL --> GR["Group rows by parent part<br/>(readableId + revision + version)"]
    GR --> RR["Batch-resolve references<br/>(items, process, work center, UOM, supplier)"]
    RR --> VAL["Validate the WHOLE file (dry run)"]
    VAL -->|"part has any error"| SK["Reject that part, report it<br/>(other parts continue)"]
    VAL -->|"part passes"| TX["Atomic transaction per part"]
    TX --> MM["Ensure make method<br/>(blank → current; explicit → find-or-create Draft)"]
    MM --> FE{"Targeted version empty?<br/>(0 materials AND 0 operations)"}
    FE -->|"no"| SK2["Skip — never clobber"]
    FE -->|"yes"| W["Write materials → operations → steps / tools / parameters"]
```

### What gets written

```mermaid
flowchart TD
    MM["makeMethod<br/>(versioned recipe, Draft)"]
    MM --> BOM["methodMaterial<br/>(BOM line, positional order)"]
    MM --> OP["methodOperation<br/>(BOP op, ordered by Op No)"]
    OP --> ST["methodOperationStep"]
    OP --> TL["methodOperationTool"]
    OP --> PA["methodOperationParameter"]
    CI["component item"] -. "derives" .-> BOM
    BOM -. "methodType · sourcingType · materialMakeMethodId · storageUnitIds" .-> CI
```

### Load-bearing decisions

- **Atomic per parent part.** Validate the whole file first; reject a part's entire recipe on any line
  error; other parts still import. In the combined file, a new part whose recipe fails is **never
  created** (we validate up front and never commit, rather than insert-then-delete).
- **Create-only / fill-if-empty, per version.** A version is filled only if it has zero materials **and**
  zero operations. A populated version is skipped and reported — never overwritten. Re-running a file is
  safe. (The empty Draft v1 the trigger creates for every new Part/Tool counts as empty.)
- **Versioning.** Blank `Make Method Version` targets the part's current method; an explicit version
  finds-or-creates that version as a **Draft** and never changes which version is Active.
- **Parent must be makeable** (Part or Tool). A Material/Consumable parent errors clearly.
- **No auto-create of references.** Missing component / process / work center / tool / UOM hard-fails
  that part's recipe only.
- **Method type & sourcing are item-level, not line-level.** A BOM line mirrors its component item for
  `methodType` / `sourcingType` / `materialMakeMethodId` / `storageUnitIds`. Because the edge function
  writes directly (it does not call `items.service.ts`), this derivation — owned by `upsertMethodMaterial`
  (#903) — is replicated inside the engine.

### Files changed

- `apps/erp/app/modules/shared/imports.models.ts` — replace the abandoned `methodMaterial` stub with
  `bom` / `operations` / `partWithMethod` field mappings, Zod schemas, and the `parts` permission.
- `packages/database/supabase/functions/import-csv/method-import.ts` *(new)* + `index.ts` — the engine:
  classify → group → resolve → validate → atomic-per-part write.
- `apps/erp/app/modules/items/ui/{Parts/PartsTable,Tools/ToolsTable}.tsx` — surface the three options.
- `apps/erp/app/modules/items/ui/Item/BoMExplorer.tsx` — re-point its (disabled) import modal off the
  removed `methodMaterial` key onto `bom`.

### Out of scope

A per-part "Import Method" button on the method page; an import-shaped BOM/BOP **export**; **overwrite**
of a populated recipe; creating a new version as a copy of a prior one; auto-creating referenced
records; non-makeable parents; non-CSV formats.

### Testing

Verified end-to-end against a real catalog: combined file creating a new part with a multi-level BOM and
mixed Inside/Outside operations + steps/tools/parameters; focused BOM and Operations files against
existing parts; skip-if-content on re-run; atomicity (a bad line fails only its part; a new part with a
failing recipe is not persisted); makeable-parent guard; explicit version fills a new Draft without
disturbing the Active version. Component-derived `methodType` / `materialMakeMethodId` confirmed in the
DB to match the component items.


jeebeez:
This is sick. Is there a demo, this could also help us improve the existing import. Would love to help with this. Do you have a demo for this

sidwebworks:
@sidgaikwad @jeebeez are there any blockers on this one?

jeebeez:
I've added some stuff and pointed some stuff. Lets make sure we are testing this one throughly

jeebeez (approved):

PR #913, opened 2026-06-19
Demo data and and resusable enviroments

jeebeez:
## What does this PR do?

- Demo data and and resusable enviroments

## How should this be tested?

- Tryall the three paths in the onboarding flow
- Try import and exporting a backup
- Use claude to check the correct importing of a backup


packages/database/supabase/functions/seed-company/index.ts
```diff
@@ -61,6 +61,28 @@ serve(async (req: Request) => {
     if (company.error) throw new Error(company.error.message);
     if (!company.data) throw new Error("Company not found");
 
+    // Idempotency guard. The whole seed runs in a single transaction, so a
```
jeebeez, line 64: good catch

sidwebworks: yeah just making some changes if you don't mind

jeebeez: yeah just take a pull before you push there should be new changes

PR #924, opened 2026-06-21
Exclude inactive customers from selection dropdowns

chasefostermfg:
## What does this PR do?

This PR prevents inactive customers from appearing in customer selection inputs (single and multi-select dropdowns) while keeping them visible in other contexts like table filters and customer avatars.

**Changes:**
- Added `customerStatusId` to the customer store and realtime data provider to track customer status
- Created `useInactiveCustomerStatusId()` hook to resolve the built-in "Inactive" status ID
- Updated `Customer.tsx` (single select) and `Customers.tsx` (multi-select) to filter out inactive customers from selection options
- Updated `RealtimeDataProvider.tsx` to fetch and sync `customerStatusId` for all customers
- Updated customer store type to include `customerStatusId` field
- Updated schema documentation to explain the filtering behavior

**Behavior:**
- Inactive customers are excluded from new-record selection inputs only
- All other statuses (Active, Lead, On Hold, Cancelled, custom) remain selectable
- Customers with no status assigned remain selectable
- The store still holds all customers (including inactive), so existing record filters and customer avatars continue to show inactive customers as intended

## How should this be tested?

1. Create a customer and set its status to "Inactive"
2. Open a form with customer selection (e.g., create a sales order or quote)
3. Verify the inactive customer does NOT appear in the dropdown
4. Verify other customers with any other status (or no status) DO appear
5. Verify that in existing record tables (e.g., SalesOrdersTable), the inactive customer still appears in filter dropdowns
6. Verify that CustomerAvatar still resolves inactive customers correctly

## Mandatory Tasks

- [x] Automated tests are in place (realtime data provider tests cover the new field sync)

https://claude.ai/code/session_01NYMCmQeu1LQNuwmhUfStgD


chasefostermfg:
@barbinbrad can you review this and make sure it's good to ship?  i tested functionality and it seems to work, but want you to think through any possible implications.

the main reason i did this because every time i go to show a demo in the ACME environment, it shows "ACME South America" as a customer...  I'm not sure how they even became a customer in there, so I've made them inactive, yet they were still showing up.

jeebeez:
I think its usually better to have a disabled UI for those or customers usually end up getting confused like why am I not seeing them. In this cased that option will just be greyed out.

Happens quite often whenever there is a UserSelect or something like this. But overall good to not let them select these

barbinbrad:
We should gray out if they're inactive, not filter.

PR #931, opened 2026-06-22
feat(import): clearer field help, undoable value mappings, per-row errors + correct counts, unit cost

jeebeez:
Import UX improvements from customer (Chase) feedback. Import is one of the first things a new customer does, so clarity matters.

> Note: this branch is built on top of #912 (data-driven creatable-form permissions + value-mapping column alignment), which isn't merged yet, so that commit is included here too. Merge #912 first or treat this as superseding it.

## What changed
1. **Explain "Default"** — an info tooltip on the value-mapping *Default* row: it's the value Carbon applies when a cell is empty or doesn't match any value above. (Customers couldn't tell what it meant.)
2. **Undoable value mappings** — each value combobox is now clearable (native ×); fixed the `onChange` that silently dropped clear events, so a mis-pick can be reset and the Default can be set back to "leave blank".
3. **Tracking Type help** — replaced the stale 2-value "tracked as inventory or not" (duplicated across all 5 item types) with one shared description enumerating **Inventory / Non-Inventory / Batch / Serial**.
4. **Clear duplicate-import error** — re-importing an item with an existing Unique ID used to abort with an opaque edge-function 500. Now a pre-flight check reports a precise, actionable per-row reason ("...already exists. Change the Unique ID...") in the existing skipped-rows panel and skips just that row so the rest import. Item/material validation failures are also reported instead of being silently dropped.
5. **Fix "Imported 0, Updated 0"** — the item branch never incremented the counters; it does now.
6. **Unit Cost** — optional field on all item imports (part/material/tool/fixture/consumable), written to `itemCost.unitCost`.

## Files
- `apps/erp/app/components/ImportCSVModal/FieldMappings.tsx` — tooltip + clearable (1,2)
- `apps/erp/app/modules/shared/imports.models.ts` — Tracking Type copy + Unit Cost field/validator (3,6)
- `packages/database/supabase/functions/import-csv/index.ts` — per-row errors, counts, unit-cost write (4,5,6)

## Notes / caveats
- The edge function is **Deno** (not covered by `tsgo`); items 4–6 there are verified by inspection against existing patterns (the customer/supplier count code; `writeItemPurchasingLeadTimes`). Runtime proof needs the dev env.
- Decision per discussion: duplicate errors are **reported, not auto-resolved** (no silent upsert of an existing item).
- A broader "validate/preview before import" redesign (the Duro-style Valid/Error/Duplicate gate) is spec'd separately in `docs/specs/import-preview-and-validation-design.md` and out of scope here.

## Test plan
- `apps/erp` typecheck: clean.
- Manual: Default tooltip shows; clearing a mapped value resets it; Tracking Type help reads the 4 values; "Unit Cost" appears as a mappable column; (edge paths) re-import a duplicate → clear skipped-row reason; successful item import shows real counts.


jeebeez:
Merging a lot fo this into sidg's work

PR #932, opened 2026-06-22
Feat/field label tooltips

aashu0148:
## What does this PR do?

Ships the field-label tooltip feature: an info icon next to selected form labels that opens a one-sentence glossary definition, plus an optional "Learn more" link to the docs page.

### Architecture (the main change)

- **New `@carbon/glossary` package.** Pure data, types, and helpers; zero runtime deps. We moved `docs/lib/glossary.ts` here and deleted the original so the docs site, the ERP, the MES, server code, and edge functions all read the same definitions instead of drifting copies. It stays React-free on purpose so a Next.js docs build, a server loader, or an edge function can import it with no UI baggage.

- **New `@carbon/field-help` package.** The React layer (one component, `LabelWithHelp`). Depends on `@carbon/glossary` and `@carbon/react`. Separate from `@carbon/glossary` because adding React to the glossary would force the docs site to take it too, and separate from `@carbon/form` because both the ERP and the MES need it and we wanted one home, not a copy in each app.

- **`TermId = keyof typeof terms`.** Every consumer passes `termId="<slug>"` typed against this union, so typos are a compile error instead of a silent missing tooltip.

- **`termId?: TermId` prop on every `@carbon/form` field.** The call site is one prop, e.g. `<Input termId="costing-method" label="Costing method" />`. Screen-reader coverage rides the existing `<FormLabel>` association, so no extra wiring at the call site.

- **Docs `<Term>` popover keeps its own renderer.** Coupling the docs UI to a React Router component would force the two stacks to share React versions and styling; the shared data source is enough and a future `chore/glossary-ui-shared` can revisit if drift bites.

### Rollout

- Wired `termId` on ~95 fields across 9 ERP modules (items, accounting/invoicing, sales, production, resources, inventory, purchasing, quality, settings) wherever a label matches an existing glossary term.
- Plumbed `termId` through 4 form components that didn't have it yet (`NumberControlled`, `InputControlled`, `ConversionFactor`, `SequenceOrCustomId`) so qualifying fields using these also get the icon.
- No new glossary entries, no label text changes, MES untouched.

---

https://github.com/user-attachments/assets/44aa8e96-cbee-4d77-bc2f-1ba80d302086

---

---

## How should this be tested?

- **Environment variables:** N/A.
- **Minimal test data:** Any seeded company with an Item, a Job, a Purchase Order, and a Fixed Asset class.
- **Expected happy path:** Hovering or focusing the info icon next to an annotated label (e.g. *Costing method*, *Replenishment System*, *Operation*, *Cost Center*, *MACRS Property Class*) opens a tooltip with the glossary definition, plus a "Learn more" link when the entry has an `href`. Esc, blur, or pointer-leave closes it.
- **Other notes:** Verified with `pnpm --filter erp typecheck` and `pnpm --filter @carbon/form typecheck`, both green. Vitest on `@carbon/glossary` covers slug shape, uniqueness, and alias uniqueness.


sidwebworks (approved):

sidwebworks:
Based on the slack conversation - please make a follow up PR to roll out this feature to the rest of the modules and implement the changes @barbinbrad mentioned in the screenshots.

barbinbrad:
@aashu0148 looks like we merged this without fixing the AI slop. can we get a fast follow on this to clean up things like the sales rfq date, and sales rfq location which aren't accurate.

PR #935, opened 2026-06-22
Feat/global empty state redesign

aashu0148:
## What does this PR do?

- Adds a `FieldEmptyState` primitive in `@carbon/form` so required dropdowns no longer render a blank picker when the option list is empty, replacing the picker with an inline "No X yet → Add X" message that is reviewer-discoverable without a click.

- Plumbs an `emptyMessage` prop through every form-layer picker wrapper (`Select`, `Combobox`, `CreatableCombobox`, `MultiSelect`, `CreatableMultiSelect`) using a wrapper-swap (when `options.length === 0`, render the empty state inside the existing `FormControl` chrome), so `@carbon/react` stays untouched and the form grid layout stays stable.

- Centralises copy, route, and permission gating in `apps/erp/app/components/Form/emptyStates.tsx` via a single `useEmptyState(entity, { onCreate })` hook backed by a per-entity config map, so adding a new entity is one row instead of one component, and users without create permission see a read-only "ask an admin" message instead of a dead CTA.

- Wires the hook into 23 domain selectors (Supplier, Customer, Location, CostCenter, Department, Process, Shift, WorkCenter, Employee, PaymentTerm, ShippingMethod, Ability, AssetClass, Part, Tool, Service, Procedure, StorageType, MaterialType, ItemPostingGroup, plus type/status variants) and migrates `ItemPurchasingForm` off its inline JSX block onto the shared primitive (Case B, custom "Add a supplier part" copy).

- Runs `lingui:extract` so all new strings land in every locale `.po` file in this PR (no untranslated keys left after merge).

---

https://github.com/user-attachments/assets/ea8a92a1-fdda-4267-9cde-190b41a33590

---

---

## How should this be tested?

- **Environment variables:** N/A.
- **Minimal test data:** A tenant with zero rows for at least one wired entity (e.g. zero Suppliers, zero Locations), and a second test user whose role lacks create permission for that module.
- **Expected happy path:** Open any form with `<Supplier name="..." />` on the empty tenant: the picker is replaced with "No suppliers yet" + "Add supplier" action; clicking opens the existing create modal (or deep-links to the route when no modal exists); after create, the picker reappears with the new option. Repeat for Customer, Location, CostCenter, Item variants, and `ItemPurchasingForm`'s filtered Case B (`allowedSuppliers` empty for a Part).
- **Other notes:** Sign in as the permission-downgraded user and confirm the CTA is replaced with the "Ask an admin with X permission" copy; verify the form grid cell height does not reflow when the picker swaps to the empty state.


aashu0148:
i had another design in mind where we show the create button as a "option" in the select box but that was a 2 click process, so i went ahead with this one. 
if this feels like taking too much space then we can explore the earlier option too. (its not a big change)

@barbinbrad

barbinbrad:
> i had another design in mind where we show the create button as a "option" in the select box but that was a 2 click process, so i went ahead with this one. if this feels like taking too much space then we can explore the earlier option too. (its not a big change)
> 
> @barbinbrad

This seems a little heavy handed and bulky to me. Can we try the other design?

aashu0148:
@barbinbrad 
image
image

updated the design to make it more minimal

barbinbrad:
I like this better personally, but what is the deciding factor for when to use this or not? For instance I was chekcing out Item and it didn't seem to work. Same with purchasing contacts and locations.

Screenshot 2026-06-22 at 11 47 12 PM

Screenshot 2026-06-22 at 11 49 09 PM

packages/form/src/components/FieldEmptyState.tsx
```diff
@@ -0,0 +1,38 @@
...
+      )}
+    >
+      {icon && <div className="mb-2 text-muted-foreground">{icon}</div>}
+      <p className="text-sm font-medium text-foreground mb-1">{title}</p>
+      <p className="text-xs text-muted-foreground max-w-[240px] leading-relaxed">
+        {description}
+      </p>
+    </div>
+  );
+};
```
barbinbrad, line 42:
```suggestion

export const fieldEmptyStateLinkClassName =
  "font-medium text-primary underline decoration-primary/30 underline-offset-2 transition-colors hover:decoration-primary";

const FieldEmptyState = ({
  title,
  description,
  icon,
  className
}: FieldEmptyStateProps) => {
  return (
    <div
      className={cn(
        "flex flex-col items-center justify-center gap-3 px-5 py-7 text-center",
        className
      )}
    >
      {icon && (
        <div className="text-muted-foreground/70 [&_svg]:size-5 [&_svg]:shrink-0">
          {icon}
        </div>
      )}
      <div className="flex flex-col gap-1">
        <p className="text-sm font-medium text-foreground text-balance">
          {title}
        </p>
        <p className="mx-auto max-w-[15rem] text-[0.8125rem] leading-5 text-muted-foreground text-pretty">
          {description}
        </p>
      </div>
    </div>
  );
};
```

barbinbrad: Just a thought with the /ui skill -- here's how that looks:

Image

aashu0148: i like the design with the "title and description", it feels more communicating. wdyt? 

It was not propagated everywhere, only at the places which were missing a create option (where options.length === 0 - mention in pr description), i should have rather added it in all cases, will update that. 

Q: should i continue this convo in PR thread or slack thread (notifications)? asking for future purpose as well.

barbinbrad (approved):

PR #936, opened 2026-06-23
bug fixes

aashu0148:
## What does this PR do?
  bug fixes:
  - Customers can no longer remove line items from a sales quote sent to them, so they can't accidentally delete lines the
  salesperson sent.
  - On a new sales-order line, the Method field now stays set when you toggle the Item / Asset switcher, instead of going
  blank.
  - The tooltip on Jobs/item-master left-panel closes before you can hover over it. 
  - Long row text in the left panel ends with a `…` instead of getting cut mid-character, and the right-side version pill no
  longer gets clipped when the panel narrows. Same fix applied to the Jobs panel and the item-master panel.
  - Sales pages no longer crash on local dev with a "cannot access before initialization" error; production was unaffected.

### DIFF
--- a/apps/erp/app/modules/shared/shared.models.ts
+++ b/apps/erp/app/modules/shared/shared.models.ts
@@ -158,6 +158,19 @@ export const sourcingType = [
   "Ship from Inventory"
 ] as const;
 
+export const taxExemptionReasons = [
+  "Resale",
+  "Government",
+  "Nonprofit",
+  "Agriculture",
+  "Industrial",
+  "Export",
+  "Medical",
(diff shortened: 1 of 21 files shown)


apps/erp/app/modules/shared/shared.models.ts
```diff
@@ -158,6 +158,19 @@ export const sourcingType = [
   "Ship from Inventory"
 ] as const;
 
+export const taxExemptionReasons = [
```
jeebeez, line 161: nice

jeebeez (approved):

PR #945, opened 2026-06-25
better inline editing UX

aashu0148:
## What does this PR do?

Adds clear edit affordances to inline-editable surfaces in Purchasing, Production, and Sales (users previously couldn't tell editable values apart from read-only text).

- Editable Table/Grid cells now show a hover pencil and enter edit mode on a single click with the value pre-selected (removes the old two-click, no-affordance flow); `EditableText`/`EditableNumber` drop a stray focus shadow.
- Quote and Supplier-Quote line pricing cells render the value as text with a hover pencil that reveals the input on click, via a new shared `EditableNumberCell` (replaces always-on borderless inputs that looked static).
- Supplier-interaction notes get a hover pencil (shown only when notes have text, hidden while focused) through a new shared `EditableNotes` component that de-dupes the editor/read-only prose pattern.

---

### Before

https://github.com/user-attachments/assets/78aee05a-10f5-4e0d-abd7-81d172179e0a

### After

https://github.com/user-attachments/assets/f7541ee8-f0ee-459e-9444-70a2f92abc38

---

---


aashu0148:
ps: i noticed while recording the video, the quantity cell in the table is highlighted while its not even editable (same for other cells too), fixed it a later commit :)

barbinbrad:
hmm. i don't know especially with the notes, it seems like too much friction. there's no edit button in notion or linear for instance.

aashu0148:
@barbinbrad removed it from notes

barbinbrad (approved):

PR #961, opened 2026-06-26
fix(harness): increase behavior gate limits (40→80 turns, $3→$5)

carbon-agent:
The behavior gate needs to log in, navigate, create/find test data, resize viewport, verify the fix, and capture screenshots. 40 turns / $3 is not enough for complex UI flows on a fresh database — claude hits `error_max_turns` with 0 screenshots taken.

Increase to 80 turns / $5 budget.

Also includes the migration deadlock retry fix from #960.

### DIFF
--- a/packages/harness/src/runner/types.ts
+++ b/packages/harness/src/runner/types.ts
@@ -121,6 +121,6 @@ export const DEFAULT_CONFIG: Omit<RunnerConfig, "cwd" | "ledgerPath"> = {
   doerMaxBudgetUsd: 5,
   judgeMaxTurns: 20,
   judgeMaxBudgetUsd: 2,
-  behaviorMaxTurns: 40,
-  behaviorMaxBudgetUsd: 3
+  behaviorMaxTurns: 300,
+  behaviorMaxBudgetUsd: 15
 };


barbinbrad:
Let's make it more like 300 for behavior. We're on a max plan. Also remeove the changes from the migraitons.ts they are already on main. And fix the type lint checks that are broken.

carbon-agent:
🤖 Addressed the review — rebased onto current `main` and force-pushed:

1. **Behavior limit → 300.** `behaviorMaxTurns: 40 → 300`. Also lifted `behaviorMaxBudgetUsd: 3 → 15` so the *turn* ceiling (not the dollar cap) is the binding limit at 300 turns — otherwise a 300-turn run would still die at ~$3. Shout if you'd rather pin the budget elsewhere.
2. **migrations.ts removed.** The branch was stale and still carried its own deadlock-retry commit; main already has the refined version (#960). Rebasing onto current `main` drops it — this PR is now a **single-purpose** harness config change.
3. **Type-lint fixed.** The failing checks came from the stale branch state (deleted packages from an old base + an unused `@ts-expect-error` in the old migration commit). After the rebase the diff is 2 lines in one typed config object, so those break sources are gone. CI will re-run to confirm.

Diff is now just `packages/harness/src/runner/types.ts` (+2/-2).

PR #965, opened 2026-06-27
perf(rls): wrap auth.uid() in (select ...) for per-statement evaluation

### LINKED ISSUE #964: RLS perf: wrap auth.uid() in (select ...) for per-statement evaluation
Hi, and thanks for carbon. I ran a static RLS analyzer ([pgrls](https://github.com/pgrls/pgrls)) over `packages/database/supabase/migrations` and found **14 RLS policies that call `auth.uid()` per row** in their `USING`/`WITH CHECK` -- 14 policies across the schema (e.g. the document SELECT/DELETE policies).

Postgres re-evaluates a bare `auth.uid()` in a policy **once per row**. Wrapping it in a scalar subquery -- `(select auth.uid())` -- makes it an InitPlan the planner evaluates **once per statement** and caches. It's predicate-equivalent (row visibility unchanged) and the optimization [Supabase documents](https://supabase.com/docs/guides/database/postgres/row-level-security) for RLS.

I've opened a PR with the wrap (a new migration) -- it's `pgrls fix --rule PERF001` output, verified to clear the finding with nothing else changed.

dmitrymaranik:
Closes #964.

The RLS policies call `auth.uid()` directly in `USING`/`WITH CHECK`, which Postgres re-evaluates **once per row**. This wraps each call in `(select auth.uid())` so the planner evaluates it **once per statement** (an InitPlan) and caches it -- the [Supabase-documented](https://supabase.com/docs/guides/database/postgres/row-level-security) RLS perf pattern.

- **Predicate-equivalent** -- `(select auth.uid())` returns the same value, so row visibility is unchanged.
- Verified with [pgrls](https://github.com/pgrls/pgrls): the PERF001 findings clear to 0 after the change, with nothing else altered.
- Added as a new migration (`20260628000000_wrap_rls_perf_initplan.sql`); the SQL is `pgrls fix --rule PERF001` output.

Happy to adjust the migration filename/placement to your conventions.


barbinbrad:
Thanks Dmitry! Have you tested this locally?

dmitrymaranik:
Yes -- I verified it on a throwaway Postgres 16 before opening this: applied carbon's full `packages/database/supabase/migrations` history, ran pgrls, applied this migration, and confirmed the PERF001 findings clear to 0 with nothing else changed. The `(select ...)` wrap is predicate-equivalent, so row visibility is identical -- it only changes when Postgres evaluates the auth call (once per statement vs once per row).

One improvement while I was here: pgrls just gained `WITH CHECK` coverage upstream, and re-checking carbon surfaced 4 more -- the write-side `auth.uid()` calls in `INSERT`/`ALL` policies (e.g. on `public.document`). I've updated this PR to wrap those too, so it now covers all 18 (14 `USING` + 4 `WITH CHECK`), re-verified PERF001 -> 0 on PG16. Happy to adjust the migration filename/placement to your conventions.

barbinbrad:
Thanks, the only thing I would ask is that you replace the `0000` from the migration name because if there's a collision on the timestamp the migraitons fail (horrible design).

dmitrymaranik:
Good call -- renamed to a real timestamp (`20260628085927_wrap_rls_perf_initplan.sql`) so it can't collide with another migration sharing the all-zeros default. Same content otherwise. Thanks again!

dmitrymaranik:
Thanks for merging, and for the careful review — both the `WITH CHECK` and the migration-naming catches made it better. This came out of [pgrls](https://github.com/pgrls/pgrls), an open-source Postgres RLS linter/verifier; if it looks useful for keeping carbon's RLS tight, a star would help other folks find it. Happy to help wire it into CI if you'd ever want regression coverage on the policies.

PR #977, opened 2026-06-29
Feat/better table exports

aashu0148:
## What does this PR do?

- Adds two per-column `meta` hooks to the shared ERP table: `exportValue` (custom CSV value for a column) and `sortBy` (server-sort key override), so CSV export and sorting no longer depend on the raw accessor. Previously composite/derived columns exported raw ids or `[object Object]`, and id-backed columns like Supplier Type sorted by UUID instead of the readable name.
- Centralizes the column-map logic into `buildColumnMaps` in `Table/utils.ts` (accessors + exportValues + sortKeyToLabel) with unit tests, replacing the inline reduce in `Table.tsx` (one place to reason about export/sort/header derivation).
- Adds a `serializeForCsv` guardrail in `Download.tsx` so a stray object value is JSON-stringified instead of shipping `[object Object]` or exploding into extra columns.
- Wires `exportValue` into many tables (Timecards, Pricing Rules, Sales Orders, Suppliers, API Keys, planning/projection, etc.) so exports show the same human-readable text as the on-screen cell, using shared formatters so cell and export never drift.
- Recreates the `salesInvoices` view to add `paymentTermName` (verbatim prior definition plus one appended column), so the Payment Method column renders and exports instead of showing blank.
- Removes a duplicate `dueDate` column in `ActionsTable` and dead `columnVisibility` keys in `SalesOrdersTable` (cleanup surfaced while wiring exports).
- Adds the new UI strings and their translations across all locales, and updates the `table-csv-export` rule doc to match the new flow.

---

https://github.com/user-attachments/assets/9c3f96ef-3c0d-4669-873d-7feac61a26f6

---


barbinbrad:
Does this work with compound `accessorKey`s? I don't see too many, just in CustomerAccountsTable and SupplierAccountsTable. We can probably deprecate that behavior if it's easier.

barbinbrad (approved):

carbon-agent (commented):
Good catch on compound `accessorKey`s — they won't export correctly as-is. The fallback path in `Download.tsx` does a flat lookup (`row["user.firstName"]` → `undefined`), so it won't walk the nested path to `row.user.firstName`.

Two options:

1. **Quick fix:** Add `meta.exportValue` functions to CustomerAccountsTable and SupplierAccountsTable (e.g. `(row) => row.user?.firstName`) — consistent with how the PR handles every other derived column.

2. **Deprecate compound accessorKeys** — only two tables use them, and flattening the data shape at the loader/query level is cleaner long-term.

Option 2 is the right call architecturally, but option 1 is a five-minute patch to unblock this PR now with deprecation as a follow-up.

PR #987, opened 2026-06-30
Download link supported in exported CSV fro documents table

aashu0148:
## What does this PR do?

- Exporting the Documents table to CSV now adds a "Download Link" column: each file gets a public URL that anyone can open to download that file directly, without logging into Carbon (so the file can be shared with people who don't have an account).
- The links are self-revoking, so there's no expiry to manage: permission is re-checked live every time a link is opened, so a link automatically stops working the moment the file is deleted or the person it was generated for loses access to it.
- Opening a broken link (tampered, deleted file, or revoked access) shows a friendly error page explaining why, instead of a raw error.
- The link only appears inside the exported CSV, not as a visible column in the table.
- Added a reusable "export-only column" mechanism so any other table can offer the same CSV download links later, not just Documents.

_the link carries a signed, tamper-proof token that encodes who it was made for; authorization is enforced live via existing row-level security, not stored in the link._

---

https://github.com/user-attachments/assets/3214bdaf-5264-4489-983b-13dfb620ed69

#### Not found
image

---


barbinbrad (commented):
Do you think this is a security vulnerability. The tokens don't seem to expire. It seems to me that if a person gets a link to a download, there's no way to revoke the ability. So if I was an employee, I might pull a list of files, then if I'm fired, I can download all the files I had access to.

aashu0148:
> Do you think this is a security vulnerability. The tokens don't seem to expire. It seems to me that if a person gets a link to a download, there's no way to revoke the ability. So if I was an employee, I might pull a list of files, then if I'm fired, I can download all the files I had access to.

@barbinbrad the token contains information of the user who downloaded the file and the document itself, we check for rbac and file existence while downloading. So if a employee is removed and their access is revoked then they can no longer download the file since they don't have the permission to access that document. 
I intentionally did not add a expiry date to the token because if someone tries to download the files after 1 or 2 months and gets that the file expired then its a bad experience for them

barbinbrad:
> > Do you think this is a security vulnerability. The tokens don't seem to expire. It seems to me that if a person gets a link to a download, there's no way to revoke the ability. So if I was an employee, I might pull a list of files, then if I'm fired, I can download all the files I had access to.
> 
> @barbinbrad the token contains information of the user who downloaded the file and the document itself, we check for rbac and file existence while downloading. So if a employee is removed and their access is revoked then they can no longer download the file since they don't have the permission to access that document. I intentionally did not add a expiry date to the token because if someone tries to download the files after 1 or 2 months and gets that the file expired then its a bad experience for them

thanks anshul. makes sense to me. @carbon-agent do you see any security problems with this approach?

carbon-agent:
The approach is secure as implemented. A few specifics from the code:

**Token integrity:** The JWT is HS256-signed over `SESSION_SECRET` via `jose` — it's cryptographically tamper-proof, not just base64-encoded. Someone can't forge a token for another user's document or another company's tenant.

**Revocation:** The download route calls `getUserScopedClient(userId)` and re-queries the document through RLS as the encoded user, with an explicit `companyId` filter. If the employee is deactivated or their document access removed, the RLS check returns no row and the link fails — the revocation is immediate and automatic, not dependent on a timer. This is actually stronger than expiry alone, which would still allow downloads during the window.

**Cross-tenant isolation:** The `companyId` guard in the query prevents a valid token for company A's document from being used against company B's data, even if document IDs were to collide.

**Type safety:** `assertPayload` requires exactly `{ userId, companyId, documentId }`, so a session JWT or any other token signed with the same key can't be mistaken for a download token.

One thing to be aware of: if `SESSION_SECRET` is ever rotated, all outstanding download tokens (in previously exported CSVs) will become invalid — not a security problem, but worth documenting so it doesn't surprise users if a key rotation happens.

No security issues from my review — the "no expiry" decision is justified given the live RLS revocation.

sidgaikwad (approved):

PR #1002, opened 2026-07-01
feat: auto fill pdfs

aashu0148:
## What does this PR do?

Adds PDF auto-fill to the New Purchase Invoice and New Sales RFQ screens, so a supplier invoice or customer RFQ document can populate the form instead of being retyped by hand.

- Upload a PDF and the header fields fill in automatically (supplier/customer, contact, location, dates, currency, payment terms, and tax), because re-keying data that's already on the document is slow and error-prone.
- Extracted customer/supplier, contact, and location are matched to existing records; if there's no match, a create dialog opens pre-filled with the extracted values so the record can be added in one step instead of leaving the flow.
- Line items from the PDF go through a mapping dialog where each line is matched to an existing catalog item, used to create a new item, or ignored (extracted part numbers/descriptions rarely line up with the catalog on their own).
- The uploaded PDF is archived onto the record so the source document stays attached.
- Extraction runs in the background (AI reads the PDF text) and the form updates when it finishes, so the page never blocks while the document is processed.

*Also included: a shared typed contract for the extracted data, the auto-fill logic pulled out of the two form components into dedicated hooks, and two fixes: the extraction request is now permission-scoped per document type (RFQ = sales, invoice = purchasing), and a failed background-job enqueue now marks the record failed instead of leaving it stuck.*

---

https://github.com/user-attachments/assets/8651cf91-2609-4533-91ff-9f2c8c6203bc

---

---

## How should this be tested?

- **Environment variables:** An AI provider key (`OPENAI_API_KEY`), optionally `AI_MODEL` (defaults to `gpt-4o`), `AI_BASE_URL`, and `EXTRACTION_CONFIDENCE_THRESHOLD` (defaults to `0.85`). The Inngest worker must be running and realtime must be enabled for the `documentExtraction` table (both are set up by the local dev stack).
- **Minimal test data:** At least one existing supplier and one existing customer (to exercise matching), plus a sample supplier-invoice PDF and a sample customer-RFQ PDF.
- **Expected happy path:** On New Purchase Invoice, click "Auto-fill from PDF" and pick the invoice PDF. Expected: supplier, contact, location, currency, payment terms, dates, and tax fill in; any value that doesn't match an existing record offers a pre-filled create dialog; line items open the mapping dialog (map / create / ignore); the PDF is saved to the record's files. Repeat on New Sales RFQ with the RFQ PDF (customer, purchasing + engineering contacts, location, dates).
- **Other notes:** Extraction is asynchronous and the created/matched records surface via realtime, so confirm the realtime websocket is connected in the browser while testing. A user with only sales permission should be able to run RFQ extraction but not invoice extraction, and vice versa.


barbinbrad:
Two requests. 

1. can we make the file upload at the bottom with the `FileDropzone` component. We don't need to say anything about "Autofill" or whatever -- and we don't need a button. Let's just autofill if we can. It should look like this:

Screenshot 2026-07-01 at 9 18 41 AM

2. is the text red? if so, can we get rid of that?

barbinbrad (changes requested):
can we use the file dropzone component and put it at the bottom of the forms? also if the text is red, can we get rid of it.

aashu0148:
@barbinbrad updated 

https://github.com/user-attachments/assets/e8776892-9b7c-41ec-8024-c246d72051c5


new auto-fill logic: it only auto fills if the detected value already present in the available options

packages/jobs/src/inngest/functions/extraction/extract-document.ts
```diff
@@ -0,0 +1,348 @@
...
+        const result = await generateText({
+          model,
+          maxRetries: 5,
+          messages: [
+            {
+              role: "user",
+              content: `${systemPrompt}\n\nFormat instructions:\n${schemaDescription}\n\nCandidate records to match against:\n${candidatesSection}\n\nHere is the text extracted from the PDF document:\n\n${pdfText}`
+            }
+          ]
+        });
```
barbinbrad: we should be using generateObject here with a schema instead of defining the schema in the prompt

PR #1007, opened 2026-07-01
Feat/plm foundation

aashu0148:
## What does this PR do?

- Lays the database foundation for the upcoming Change Order (ECO) / PLM feature so the later feature PRs build on stable, shared tables instead of each shipping its own schema.
- Ships data only and changes no existing behavior: the new tables and columns are inert until a feature PR wires them up (verified nothing reads or writes them yet).
- Adds the ability to track an item's revision stage (Design/Prototype/Production/Obsolete) and a per-company release-control setting (off/warn/enforce), both defaulted so existing data is unaffected.
- Access to change orders is gated on the same Parts permission as items themselves, since change orders live under the Items area (the original clone used Production permissions, which would have let someone create a change order they then could not view).


apps/erp/app/modules/items/index.ts
```diff
@@ -1,3 +1,4 @@
+export type * from "./changeOrder.types";
```
barbinbrad: These should go in items.models.ts

barbinbrad (changes requested):
Looks great!

But I'd rather not merge like this. We're just assuming that we know the shape of the data. But until we build it all out we really don't know. Instead of doing incremental merging for a big feature, I would rather merge one big one, so that we can continue to edit `plm-change-orders.sql` as needed. Personally I can't tell whether this data model is correct or not until we wire it all together.

aashu0148:
@barbinbrad this PR is not to merge yet, but i am planning to divide the feature into multiple PRs so its easy to take a look at each PR. 
I will change this PR to draft, just in case.

barbinbrad:
> @barbinbrad this PR is not to merge yet, but i am planning to divide the feature into multiple PRs so its easy to take a look at each PR. I will change this PR to draft, just in case.

you can try it but i think it's going to be annoying to manage a bunch of stacked PRs. we really only want to test that it works -- and verify that we're following conventions.

PR #1021, opened 2026-07-03
perf: ledger query performance — indexes, draft exclusion, snapshot + delta

barbinbrad:
## Summary

Full audit + optimization of the ledger-based read paths (`itemLedger`, `costLedger`, `journal`/`journalLine` → financial statements, on-hand quantities, COGS). Everything here keeps read latency flat as the ledgers grow, without partitioning, write-path triggers, or dual writes. Audit and design docs are in `.ai/plans/2026-07-02-ledger-query-performance.md` and `.ai/plans/2026-07-02-ledger-snapshot-delta.md`.

### 1. Missing indexes (`20260702231544`)

- `itemLedger` had **no index on `itemId`/`locationId` at all** — the `issue` edge function's per-material `SUM(quantity) WHERE itemId AND locationId` ran as a sequential scan of the entire multi-tenant table, inside the posting transaction. Now a covering index (`INCLUDE storageUnitId, quantity, trackedEntityStatus`) makes those index-only scans.
- `journalLine.journalId` (NOT NULL FK) was unindexed — hit the `journalEntries` view, cascade deletes, void/reverse reads, and the per-row Draft-check RLS.
- Plus: `itemLedger(companyId, locationId, itemId)`, document-id partials on both ledgers, `journal(companyId, postingDate)`, `costLedger(supplierId)`, and `supplierLedger` (which had nothing but its PK).

### 2. Correctness + query-shape fixes (`20260702233127`, `20260702234618`)

- **Draft journals no longer leak into financial statements.** `accountTreeBalancesByCompany` / `accountTreeBalances` / `journalLinesByAccountNumber` gate every SUM on `journal.status <> 'Draft'`. Reversed journals still count (a reversal offsets the original, it doesn't remove it). The account drill-down (#1018) now excludes Draft too, so it still ties out with the balances.
- **Halved the ledger scans for translated/consolidated statements.** `translateCompanyBalances` now translates the balances the caller already computed, using a new rates-only `getConsolidationRates` RPC, instead of re-running the full `journalLine` scan through `translateTrialBalance`. Consolidated mode previously did 2 full scans × N companies per render; now 1 per company.
- **`get_inventory_quantities` gained an optional `item_id` parameter** — item detail pages no longer aggregate the whole location's ledger + POs + SOs + jobs to show one item.
- **Fixed a real bug in `update-purchased-prices`**: the raw-row `select("*")` on `costLedger` was silently capped at 1000 rows by PostgREST, skewing weighted-average costs for high-volume items. Now aggregates in SQL.
- `getItemCostHistory` bounded to 500 rows.

### 3. Snapshot + delta (`20260702235112`, `20260702235521`, `20260702235858`)

The structural fix for "every balance is a SUM since inception":

- **GL**: new `accountingPeriodBalance` table stores cumulative per-account balances, written by `snapshotAccountingPeriodBalances(companyId, periodId, userId)` — to be called by the period-close flow (period-closing branch) inside its close transaction. `accountTreeBalancesByCompany` becomes *snapshot + open-period delta*, pruning the `journalLine` join to bounded `postingDate` ranges via the new `journal(companyId, postingDate)` index. With no snapshots it takes the identical full-scan path, so **this PR changes no GL behavior** beyond the Draft exclusion; the win activates when period closing lands. Invariants the close flow must enforce are documented in the migration header and `.ai/plans/2026-07-02-ledger-snapshot-delta.md`.
- **Inventory**: new private `itemLedgerSnapshot` matview (30-min pg_cron refresh) snapshots only the **immutable untracked** ledger rows, cut off at `createdAt < now() - 1h`. Live reads = snapshot + rows past the cutoff + **all tracked rows live**, so status-aware sums (`Rejected`/`On Hold`) are never stale and results are exact. A `createdAt` cutoff is used instead of an `entryNumber` watermark because SERIAL values don't respect commit order (an entryNumber watermark can permanently miss late-committing rows). `itemStockQuantities` (the UI item-store matview) is untouched.

## Verification

- **In-transaction equivalence suite** against a live DB (rolled back): baselines cloned via `pg_get_functiondef`, then compared with `EXCEPT` in both directions. All passed:
  - Draft journals excluded from balances (150, not 160, with a Draft fixture present)
  - No-snapshot fallback identical to the full scan
  - Cumulative snapshot value correct (drafts excluded)
  - Snapshot + delta ≡ full scan across 5 date-range shapes (wide default, straddling, before-snapshot, empty window)
  - Fresh postings after close picked up by the delta
  - Matview holds only pre-cutoff untracked rows; tracked-entity status flip to `Rejected` visible **immediately**; single-item variant, delta postings, and the inventory value report all ≡ baseline
- `pnpm exec turbo run typecheck --filter=./apps/erp` ✅ (37–46s)
- Biome clean; `deno check` on the edited edge function shows zero new errors vs baseline
- All migrations idempotent (`IF NOT EXISTS` / `DROP IF EXISTS` guards) and applied cleanly in a rolled-back transaction

## Notes for reviewers

- The committed DB types are cloud-generated post-deploy, so the two call sites touching new SQL (`item_id` arg, `getConsolidationRates`) use narrow casts with TODO-style comments — same pattern as the tie-out RPCs.
- `accountTreeBalances` (group-wide chart of accounts) intentionally stays on the full scan — snapshots are per company.
- Deferred/dropped from the original plan (documented in the audit doc): `count: "exact"` removal on `getJournalEntries` (fragile vs view-column filters, and the new `journalId` index makes the view cheap), and removing `itemLedger` from the realtime publication (`StockMovementsTable` subscribes to it).

## Summary by CodeRabbit

* **New Features**
  * Added faster ledger balance retrieval using period snapshots and incremental updates.
  * Inventory quantity lookups can now be limited to a specific item for quicker detail views.
  * Added safeguards to keep tracked inventory statuses reflected immediately.

* **Bug Fixes**
  * Draft journal activity is excluded from ledger totals and balances.
  * Period reopening now invalidates affected balance snapshots safely.
  * Purchased-price updates now process complete historical totals without row truncation.
  * Cost history is limited to the 500 most recent entries.

* **Documentation**
  * Updated accounting and inventory guidance for the latest balance and snapshot behavior.


sidwebworks:
@coderabbitai review this

PR #1027, opened 2026-07-03
🟢 better table row select

aashu0148:
## What does this PR do?
ability to shift + select rows in a table

https://github.com/user-attachments/assets/dfaece94-3a33-4ce0-ba29-aba146de2fa1


naveenkash (approved):
lgtm

PR #1103, opened 2026-07-08
fix(erp): match step cards to tool cards and round first/last cards

sidgaikwad:
## What

In the Bill of Process editor (both the **item method** editor and the **job** editor), the step list and tool list are rounded containers, but their first/last rows weren't rounded — so the last row's square-cornered border poked past the rounded container, leaving a stray line beneath the last step. Step rows also lacked the `bg-card` background the tool rows already had, so the two lists looked inconsistent.

## Changes (CSS-only)

- Step rows now use the same `bg-card` + border as tool cards.
- First/last rows in **both** the step and tool lists get `rounded-t-lg` / `rounded-b-lg` so the group's corners sit flush inside the rounded container — removing the square-corner border line.
- Moved `bg-card` off the step-list container (it now lives on the rows, matching the tool list).

No logic changes — purely `className` tweaks in `BillOfProcess.tsx` and `JobBillOfProcess.tsx`.

## Why a separate PR

Split out from the in-progress MES assembly-view work so this small visual fix can land on its own without waiting for that feature.

## Summary by CodeRabbit

* **Style**
  * Updated bill-of-process UI list styling across steps, attributes, parameters, and tools with consistent first/last item rounding and no bottom border on the final row.
  * Ensured editable/collapsible item wrappers consistently use card-style backgrounds.
* **Bug Fixes**
  * Adjusted step preview behavior to treat empty step records as “no records,” avoiding unnecessary bordered preview containers.


aashu0148 (approved):

barbinbrad:
Looks like we missed `QuoteBillOfProcess.tsx`

sidgaikwad:
@barbinbrad done with then changes for quotebillofprocess

barbinbrad (approved):

PR #1116, opened 2026-07-09
MES: suggest batch on issue + return unused picks at job complete

sidwebworks:
## What

Two changes to the MES material issue / pick flow, plus a consumption-ledger bug fix found along the way.

### 1. Suggest the batch when issuing (no picking list linked)
When an operator issues a batch/serial material that has **no** picking list allocated, the Issue modal now pre-fills the batch/lot selection with the **same allocation a picking list would make** — FEFO/FIFO/LIFO per the item's pick method, netting out existing allocations, spilling across lots to cover the estimated quantity. It's a starting point; fully editable.

- Adds `p_sort_method` to `get_available_tracked_entities` so the single-item availability query honors the item's pick order.
- Ordering is done authoritatively in the app (`sortLotsByPickMethod`) — a `LANGUAGE sql` function's internal `ORDER BY` is not guaranteed to survive through PostgREST inlining (caught in testing; noted in `.ai/lessons.md`).
- New endpoint `api/suggested-allocation` + modal wiring.

### 2. Return unused picks to the warehouse at job complete
When a batch/serial lot is picked to the lineside shelf but not fully consumed, the remainder now returns to its warehouse source when the job completes — so inventory reflects only what was actually consumed and the stock is re-allocatable.

- New `post-picking` case `returnPickedRemainder`: walks the picked entity's split lineage and moves its remaining lineside on-hand back to source, decrementing the allocation.
- Sweep in `finishJobOperation` runs it per allocation once `job.status = 'Completed'`. Idempotent.

### Fix: consumption booked against the wrong bin
The `issue` edge function chose the Consumption/split ledger's storage unit via an arbitrary `.find` over a `createdBy`-ordered list, so consuming picked stock was booked against the warehouse bin instead of lineside — leaving a per-bin-negative, inconsistent ledger and misplacing the remainder. Now resolved to the entity's actual current bin (`resolveTrackedEntityBin`). This is what makes the return in (2) correct.

## Testing
- Unit tests for the sort + greedy-fill logic (`inventory.greedy.test.ts`, 11 cases).
- Verified end-to-end against seeded real data on the dev DB:
  - Suggestion: FEFO and FIFO produce the correct lots + quantities via the authenticated route.
  - Bin fix: consumption books at lineside, no per-bin-negative balances.
  - `returnPickedRemainder`: remainder moves lineside → source, inventory-conserving, no negatives, idempotent.

## Notes
- The return **sweep glue** in `finishJobOperation` is typechecked but not yet driven through the full shop-floor UI path (`finishJobOperation → trigger → complete → sweep`); the building blocks it calls are verified on real data.
- Scoped to **batch/serial** materials (out of scope: the "picking list linked" display, section 2 of the design).
- Generated DB types are intentionally not included — they regenerate on migrate.

## Summary by CodeRabbit

* **New Features**
  * Added automatic batch/serial “picking-list-style” allocation suggestions when issuing tracked material without an existing picking allocation.
  * Added support for returning unconsumed picked remainder stock when completing a job operation.
* **Bug Fixes**
  * Fixed tracked-item inventory ledger bookings to use the correct resolved storage unit/bin during consumption and split flows.
  * Improved allocation/availability ordering to consistently follow FEFO/FIFO/LIFO behavior.
* **Documentation**
  * Updated MES job-operation and traceability docs for the pick → consume → return flow.
* **Tests**
  * Added unit tests for allocation filling and pick-method sorting behavior.


apps/mes/app/services/operations.service.ts
```diff
@@ -186,11 +187,83 @@ export async function finishJobOperation(
...
+          trackedEntityId: allocation.trackedEntityId,
+          locationId,
+          userId: args.userId,
+          companyId: args.companyId
+        }
+      })
+    );
+  });
+
+  await Promise.all(returns);
```
sidgaikwad: functions.invoke resolves to { data, error } instead of rejecting, so Promise.all here never sees a failed return — if a returnPickedRemainder call errors, the remainder silently stays lineside and nothing's logged. Can we check each result's .error and log it via log (line 25)?

apps/mes/app/services/allocation.ts
```diff
@@ -0,0 +1,101 @@
...
+ * internal ORDER BY because a `LANGUAGE sql` set-returning function is inlined by
+ * the planner and its ordering is NOT guaranteed to survive through PostgREST
+ * (verified: FIFO's `createdAt`-only order was dropped). Stable + pure.
+ *   Default/FEFO → expiration asc (nulls last), then createdAt asc
+ *   FIFO         → createdAt asc
+ *   LIFO         → createdAt desc
+ */
+export function sortLotsByPickMethod<T extends SortableLot>(
+  lots: T[],
+  sortMethod: PickMethodSortMethod | string
```
sidgaikwad: Nit: PickMethodSortMethod | string collapses to string, so this param has no compile-time typo safety. The only caller already passes the enum from getPickOrder, so it's safe to drop | string and take PickMethodSortMethod directly.

sidgaikwad (approved):

PR #1126, opened 2026-07-10
ci: sync self-hosted functions should depend on deploy apps

### LINKED ISSUE #1125: ci: sync self-hosted functions workflow should depend on deploy apps
## Problem

The `Deploy Self-Hosted Functions` workflow (`.github/workflows/functions.yml`) currently triggers on `push` to `packages/database/supabase/functions/**` independently of the deploy pipeline. This means it can run before the app deployment has completed, which can cause sync issues if the functions depend on the newly deployed app version.

## Acceptance Criteria

- [ ] `functions.yml` is triggered by `workflow_run` on `Deploy Apps` (completed + succeeded) rather than directly on `push`
- [ ] A delay is added after the deploy completes before running the sync (e.g. a sleep step, or a `wait-for` action) to give the deployment time to fully propagate
- [ ] `workflow_dispatch` manual trigger is preserved for on-demand runs
- [ ] The path filter (`packages/database/supabase/functions/**`) behavior is preserved — only sync when those files actually changed (can check via `dorny/paths-filter` or similar since `workflow_run` doesn't support path filters natively)
- [ ] CI passes on a test branch / the workflow file is valid YAML

## Notes

- `inngest.yml` already uses `workflow_run` correctly — mirror that pattern
- The delay should be conservative (e.g. 60s) to give ECS/infra time to stabilize post-deploy
- Assigned to carbon-agent to build

carbon-agent:
## Summary

Switches the `Deploy Self-Hosted Functions` workflow from a direct `push` trigger to a `workflow_run` dependency on `Deploy Apps`, so functions only sync after a successful app deploy — no more racing ECS/infra.

## Changes

- Trigger changed from `push` → `workflow_run` on "Deploy Apps" (mirrors `inngest.yml` pattern)
- Job-level guard: only runs if `Deploy Apps` succeeded (skipped on failure)
- 60s stabilization delay after deploy (skipped for `workflow_dispatch`)
- `dorny/paths-filter@v3` preserves path-filter behavior — only syncs when `packages/database/supabase/functions/**` actually changed
- `workflow_dispatch` preserved for on-demand manual runs (bypasses path filter + delay)

Closes #1125

## Summary by CodeRabbit

* **Deployment Improvements**
  * Self-hosted function deployments now run automatically after a successful app deployment completes.
  * Added manual trigger support for function deployments.
  * Function deployment and related cleanup now run only when relevant function files change (or when manually dispatched).
  * Added a short stabilization delay before applicable function deployments.

### DIFF
--- a/.github/workflows/functions.yml
+++ b/.github/workflows/functions.yml
@@ -1,15 +1,22 @@
 name: Deploy Self-Hosted Functions
 
 on:
-  push:
-    branches: [main]
-    paths: ["packages/database/supabase/functions/**"]
+  # Run after the Deploy Apps workflow completes so functions sync only once the
+  # app deploy has finished (avoids racing ECS/infra changes).
+  workflow_run:
+    workflows: ["Deploy Apps"]
+    types:
+      - completed
+    branches:
+      - main
   workflow_dispatch:
 
 jobs:
   functions:
     name: Deploy Self-Hosted Functions
     runs-on: ubuntu-latest
+    # Only run if the triggering Deploy Apps workflow succeeded (or if manually triggered)
+    if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
     strategy:
       matrix:
         environment:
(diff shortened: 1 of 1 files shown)


carbon-agent:
Fixed in a674f99ae — added `base: ${{ github.event.workflow_run.head_sha }}~1` to the `paths-filter` step. Now compares the pushed commit against its immediate parent, so changes to `packages/database/supabase/functions/**` are reliably detected on every `workflow_run` trigger.

sidwebworks (approved):

PR #1133, opened 2026-07-12
fix: carry Customer PO, Sales Person, and Engineering Contact through new Sales Order creation

carbon-agent:
## Problem

When creating a new Sales Order directly (not via the Quote → Order conversion), three fields visible in the New Sales Order form were silently dropped and never persisted to the database:

- **Customer PO Number** (`customerReference`) — the field with the tooltip "The customer's own reference for this document"
- **Sales Person** (`salesPersonId`)
- **Engineering Contact** (`customerEngineeringContactId`)

The validator (`salesOrderValidator`) captured all three fields correctly, and the route action spread `validation.data` into `insertSalesOrder()` — but `insertSalesOrder()` didn't accept them in its input type, so TypeScript ignored them, and they were never included in the `INSERT` statement.

Note: the Quote → Order conversion path was unaffected — it goes through a separate edge function that already handled `purchaseOrderNumber → customerReference` correctly.

## Fix

Added `customerReference`, `salesPersonId`, and `customerEngineeringContactId` to:
1. The `insertSalesOrder()` input type
2. The `salesOrder` insert statement

## Testing

- Create a new Sales Order with a Customer PO Number, Sales Person, and Engineering Contact filled in
- Save — all three should now appear correctly on the Sales Order detail page
- Existing Quote → Order conversion flow is unaffected

## Summary by CodeRabbit

* **New Features**
  * Sales orders can now include customer references, customer engineering contacts, and assigned salespeople.
  * These optional details are supported when creating sales orders through the available tools.

### DIFF
--- a/apps/erp/app/modules/sales/sales.service.ts
+++ b/apps/erp/app/modules/sales/sales.service.ts
@@ -4956,6 +4956,9 @@ export async function insertSalesOrder(
     requestedDate?: string;
     promisedDate?: string;
     notes?: string;
+    customerReference?: string;
+    customerEngineeringContactId?: string;
+    salesPersonId?: string;
     customFields?: Json;
   }
 ): Promise<{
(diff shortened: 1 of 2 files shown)


chasefostermfg:
@barbinbrad tested and confirmed this is fixed.  you may want to review code as well before merging.

PR #1135, opened 2026-07-13
Manufacturing inventory fixes

sidwebworks:
## Inventory count hardening + history

**Posting & correctness**
- Post reconciles `delta = counted − systemQuantity` (preserves concurrent movements); `FOR UPDATE` lock + guarded `Pending→Posted` flip against double-post/races.
- Serial validation (qty 0/1) enforced in UI **and** server; offending lines flagged on post.
- **In-place rectify** (`Posted→Draft`) re-posts corrections linked via `itemLedger.correctionOfItemLedgerId` + `isCorrection` view flag.

**Workflow & UX**
- Edit only in Draft (no lock mode); Draft auto-enters edit mode (`Table.forceEditMode`, toggle hidden). Blind counts hide system qty until Posted.
- Click-away clears the selected cell; `EditableNumber` commits on native blur (fixes react-table desync).
- **Count History** drawer: per-post version timeline (original + rectifications), grouped by posting round. Nested corrections in Stock Movements.

**Audit & logging**
- `inventoryCount`/`inventoryCountLine` registered in audit config + trigger migration.
- `handleError` logs genuine 5xx with requestId (erp + mes); `unhandledRejection`/`uncaughtException` handlers; edge-function error bodies surfaced.

**Housekeeping**
- Filled missing translations (12 locales); merged latest `main`.

## Summary by CodeRabbit

* **New Features**
  * Added inventory count history showing original postings and rectifications, with “Rectify” and reopen flows that refresh system quantities.
  * Enhanced stock movement tables to display corrections with expandable details and clearer quantity-delta presentation.
  * MES material-issue screens now restore previously picked lots and serials.

* **Bug Fixes**
  * Strengthened serial-tracked posting validation (including clearer, line-level error feedback).
  * Improved table inline editing/expansion and click-away behavior.

* **Documentation & Localization**
  * Updated audit coverage and added count/correction terminology across supported languages.

* **Platform / Observability**
  * Improved server error handling and request correlation logging.


sidwebworks:
Addressed the review comments in `c8ee7173c`:

- **EditableNumber — re-parsing DOM text (Critical):** now parses the blurred value with locale-aware separators (`Intl.NumberFormat` group/decimal) and treats `NaN`/empty as *no value* (never persists a stray `0`). The DOM-blur read is retained deliberately — react-aria's `onChange` is dropped when moving to another cell unmounts the input mid-commit (that was the original state-desync bug), so it isn't a reliable commit path here. Locale is read hook-free from `<html lang>` since the editable cell is a nested render fn, not a hook context.
- **post-inventory-count — 400 for everything:** non-validation errors (not-found, no-longer-pending, DB/runtime) now return **500**; only line-validation stays 400.
- **Rectify not atomic:** new `rectifyInventoryCount` does lock (`FOR UPDATE`) + `Posted` guard + resnapshot + `Draft` flip in **one transaction**. The route no longer does the read-then-two-writes dance. (Also removed the now-unused `resnapshotInventoryCountLines` export — which resolves the MCP `updatedBy`-injection comment, since that tool no longer exists.)
- **Chained corrections invisible:** grouping now resolves each correction to its **ultimate on-page root ancestor**, so a rectification-of-a-rectification nests under the top-level original at any depth.
- **Fatal process handlers:** `uncaughtException` now `process.exit(1)` after logging (erp + mes) so the supervisor restarts a clean worker. `unhandledRejection` stays log-only — a stray rejection is usually recoverable and crashing the whole server on each one is worse.
- **itemLedgers view companyId joins:** added `companyId` predicates to the `location`, `storageUnit`, and `trackedEntity` joins (new migration `20260713130718`). `modelUpload` is reached via `item.modelUploadId`, i.e. already company-scoped through the `item` join.
- **`material.id` optional chaining:** fixed the picked-allocation hook call to `material?.id`.

**Deferred (not fixed here) — `insertManualInventoryAdjustment` atomicity:** this is a legitimate gap, but it's function-wide (~8 sequential supabase writes across several branches), not specific to the two lines flagged, and predates this PR. A correct fix is an end-to-end Kysely conversion of a hot inventory path; wrapping only the flagged pair would give false confidence. Tracking it as a separate focused change rather than bundling a risky partial fix into this PR.

PR #1151, opened 2026-07-15
feat: finite-capacity scheduling — capacity reservations, ability-gated operators, explainable Gantt

naveenkash:
## What this does

Finite-capacity production scheduling, end to end — plus the planning-engine
extraction and the ERP surfaces that sit on top of it.

### Scheduling engine (now in-process, not an edge function)

- The finite scheduler and MRP run **in-process in Node** via
  `@carbon/ee/planning` (`runLocationSchedule`, `runMrp`, `runExpediteWhatIf`).
  The `schedule` and `mrp` Deno edge functions were **deleted** — no cold start,
  no HTTP hop. The pure MRP compute engine stays in the edge-lib
  (`@carbon/database/mrp-engine`) and is shared.
- One deterministic **forward-ASAP whole-location pass**. Work centers are finite
  (capacity 1 — one op at a time). For ability-gated processes, qualified
  **people** are finite too: work only accumulates while a qualified employee is
  on shift and unreserved. Placements persist as `capacityReservation` rows
  (`WorkCenter` + `Employee` kinds; `OperatorPool` is read-tolerated legacy).
- **Work-center operating hours** via an availability ladder: `alwaysOn`
  (lights-out 24×7) → explicit `workCenterShift` rows → the location's shifts →
  a stock Mon–Fri 8h week. Open maintenance dispatches flagged
  `takesWorkCenterOffline` subtract their window from the schedule.

### Dual dates — due (need-by) vs projected

- Every op carries a backward, demand-anchored **need-by** target
  (`jobOperation.dueDate`, from `computeNeedByDates`) and a forward
  **projection** (`startDate` + `jobOperation.projectedCompletionAt`). Targets
  never floor placement; an op projected past its need-by shows an informational
  **amber behind-target** state (BOP rows, ops-board cards, MES op detail), and a
  late job's cause names the first behind-target op.
- `manuallyScheduled` pins a human-owned need-by (taken as-is, propagates
  upstream, never overwritten). Rescheduling always goes through
  `notifyScheduleInputsChanged` → the debounced wave regenerates the whole
  location; dates are never written directly.

### People / manning board

- `peopleAssignment` / `peopleAbsence` turn a staffed station into a **team**: the
  station still runs one op at a time, but the whole present team is booked on it
  and labor is parallelized across them (setup/machine never compressed). The
  board is a commitment — the any-qualified fallback may only draw on true
  floaters, so moving your only welder off a station makes that op an
  unschedulable placeholder rather than silently double-booking.
- **`location.requiresStaffing`** (per-location, default off, edited on the
  Production → Scheduling settings): when on, work is placed only where an
  operator is manned, except lights-out stations.

### Forecast board

- `/x/scheduling/forecast` (renamed from "Simulations"; the old per-job Timeline
  page was removed) — a cross-job resource-load Gantt of `capacityReservation`s,
  with non-working-hours shading from the same availability ladder the scheduler
  uses, day/week/shift navigation, and honest "unschedulable placeholder" rows.

### Plan gating

- The **Forecast** board and the **Demand Forecasts** page are gated behind the
  Business plan (`FEATURE_PLANS.FORECAST = [Business, Partner]`). Nav items stay
  visible; navigating renders a `ForecastUpgradeOverlay`. Demand-forecast
  mutations also enforce it server-side via `requirePlan`. Gating is a no-op off
  the Cloud edition, so enterprise self-hosted keeps both surfaces.

### ERP surface / polish (same branch)

- Home dashboard redesign (search, agent onboarding, dynamic greeting), Recently
  Viewed fixes, `bg-card` detail/explorer backgrounds, a **Publish** button on the
  training header, and a purchasing-planning fix that stamps the selected supplier
  onto planning orders (and re-seeds defaults when rows revalidate).

## Architecture note (changed since this PR opened)

This PR originally shipped the scheduler as a `schedule` edge function with
`capacityReservation` (WorkCenter + OperatorPool) and a `schedulingPolicy`
dispatch knob. That was **superseded** mid-branch: the engine moved in-process
into `@carbon/ee/planning`, operator pools became finite people
(`Employee` reservations + the manning board), the configurable dispatch policy
was dropped in favor of a single placement order (`jobOperation.priority`), and
qualification became presence-based (`employeeAbility`, no active/trainingCompleted
gate). The earlier self-review "known issues" list targeted the deleted
edge-function engine and no longer maps 1:1 — please re-review against the current
`@carbon/ee/planning` code rather than that list.

## Verification

- Scoped typechecks green (`erp`, `@carbon/ee`, `@carbon/jobs`, `@carbon/database`).
- Deno engine tests + erp vitest for the scheduling/need-by/placement paths.
- Dual-dates and forecast flows browser-verified against seeded scenarios
  (playbooks under `.ai/playbooks/`).
- Forecast plan-gating: non-gated render verified in the running app; the gated
  overlay reuses the shipped `UpgradeOverlay` primitives (not yet screenshotted in
  a Cloud/Community stack).

## Merge

`origin/main` merged into the branch to clear conflicts. Resolutions:
`mrp.ts` (union of main's per-company selection + our in-process `runMrp`),
`mrp-system.md` (selects from `company`, runs in-process), the generated
`salesInvoiceLocations` relationships in both `types.ts` mirrors (reorder-only
conflict, kept all three country FKs), and a regenerated MCP `tool-metadata.json`.


barbinbrad:
why do we need the active and training complete booleans in the abilities table? can't we do the same thing without them? it seems like a false premise to say that each ability requires training. and it seems silly that the active checkbox functions the same as a delete checkbox. and that if we add an employee to the ability, that employee doesn't show up in qualified employees count unless the training complete button is true.

barbinbrad:
these are too buried. no one is ever going to find them. remember the department heads who are scheduling are some of the least tech savvy people because they are experts in the work.

Screenshot 2026-08-12 at 12 55 40 AM

barbinbrad:
Can we check the capacity page for a memory leak?

PR #1153, opened 2026-07-15
bug fix: subdivide batch lots when splitting disposition rows

aashu0148:
## What does this PR do?

- Splitting a disposition line on a quality issue (NCR) now works when the flagged material is a single batch lot. Before, entering a partial quantity always failed, because the system could only shuffle whole lots between rows and never divide one.
- When you split off part of a batch, the system now creates a second lot for that quantity and leaves the rest on the original row, so each portion can take its own outcome (scrap some, rework some). This is the same lot-splitting the shop floor already does when only part of a lot is consumed, so inventory history and traceability stay consistent.
- Disposition card polish: the read-only quantity no longer looks like an editable text box, long status labels now truncate with an ellipsis instead of overflowing into the actions button, the split dialog no longer has a large gap above the first field, and the row contents are top-aligned.

#### Video Demo:

https://github.com/user-attachments/assets/b90f86c9-4001-4f23-b640-d11873ff78ff

## Summary by CodeRabbit

* **New Features**
  * Improved quality issue disposition workflows for moving tracked quantities, splitting issue items, and closing issues with inventory/status updates.
* **Bug Fixes**
  * Disposition item quantity is now read-only to prevent unintended edits.
  * Split quantity and split inputs are strictly validated with clearer destructive feedback.
  * Closing a disposition now performs a pre-check of the disposition plan before posting.
  * Direct quantity updates via the item update action are no longer accepted.
* **Style**
  * Refined disposition status rendering with consistent color mapping, truncation, and improved action-menu visibility.
* **Localization**
  * Added “Enter a quantity between 1 and {0}” validation messaging across supported languages.


apps/erp/app/modules/quality/quality-disposition.server.ts
```diff
@@ -0,0 +1,773 @@
...
+// creates a new lot for `moveQty` linked to `newRowId`, decrements the original
+// lot to `keepQty` (kept on its existing row), and writes split genealogy
+// (trackedActivity "Split" + inputs/outputs + net-zero "Batch Split" itemLedger,
+// which leaves on-hand unchanged).
+//
+// NOTE: the same lot-subdivision + genealogy also lives in the MES issue edge
+// function (packages/database/supabase/functions/issue/index.ts, case
+// "trackedEntitiesToOperation"). They can't share code across the app/Deno
+// boundary today — keep the two in sync if the genealogy shape changes.
+async function subdivideBatchEntity(
```
sidwebworks, line 195: this inserts three Batch Split itemLedger rows (`-entityQty`, `+keepQty` on source, `+moveQty` on new entity) but never sets `storageUnitId`, so they all land in a the `null` storage unit.

Storage unit on-hand is computed from `itemLedger` grouped by `storageUnitId` (picking + available-entity views). Since our adjustments sit in NULL instead of the lot's real bin, they don't cancel:

- Lot received in SU X: `+100`. Split off 30 → we write NULL-SU `-100/+70` + new lot NULL-SU `+30`.
- SU X still nets **+100** (shows full 100, not 70). The 30 sits in NULL SU, invisible to picking.

Item total stays correct - only SU/lot detail breaks, which is what picking uses.

MES already solved this: `resolve-tracked-entity-bin.ts` finds the bin holding the stock, and MES split sets `storageUnitId` on all 3 rows. NCR split needs the same.

So basically: resolve source lot's real storage unit, set `storageUnitId` on the 3 split rows.

sidwebworks (changes requested):
Left comments

Also `itemQuantityValidator` in `quality.models.ts:529` and the case "quantity" in `item+/update.tsx` both are dead code now. Should we clean it up after confirming manually?

PR #1164, opened 2026-07-17
feat(suggestions): opt-in sending suggestions to Carbon Slack

naveenkash:
## What

Adds a **"Send to Carbon"** checkbox (default on) to the suggestion popovers in ERP and MES. When checked, the submitted suggestion is posted to Carbon's own Slack workspace (`#feedback`) so the team can act on customer feedback. When unchecked, the suggestion stays entirely within the customer's instance.

The checkbox only renders on deployments that can actually forward (see `CARBON_SLACK_ENABLED` below) — self-hosted instances never see it.

## How

- **UI** (`apps/erp` + `apps/mes` `Suggestion.tsx`): second checkbox under "Submit anonymously", same local-state + `Hidden` field pattern; resets to checked after submit. Rendered only when `CARBON_SLACK_ENABLED`.
- **`CARBON_SLACK_ENABLED`** (`packages/env`): derived, non-secret browser flag — server-side it's `Boolean(SLACK_BOT_TOKEN)`; it crosses the wire via `window.env` as `"true" | ""` (keeping `window.env` homogeneously string-typed) and the token itself stays server-only.
- **Validators**: `sendToCarbon` added via the house boolean pattern (`zfd.text(...).transform(v => v === "true")`) — a missing field parses to `false`, so stale/cached clients fail closed (no post).
- **Routes**: after the DB insert, `if (sendToCarbon)` calls the new shared helper. Form-supplied identifiers are no longer trusted: `userId` is used only as an anonymous on/off signal (the row and Slack message attribute the *authenticated* user or nobody), and `attachmentPath` must live under `{companyId}/suggestions/` or it's dropped. The existing tenant-facing notification (email + Slack DMs to the company's notification group) is untouched.
- **`postSuggestionToCarbonSlack`** (new, `packages/lib/src/slack.server.ts`): posts via the env `SLACK_BOT_TOKEN` (Carbon's workspace — deliberately *not* the tenant's `companyIntegration` token, and not the `send-slack` job, which resolves the tenant token first). Same mechanism as the existing feedback widget.
  - Returns early when no `SLACK_BOT_TOKEN` is configured (self-hosted), even if a forged form submits `sendToCarbon=true`.
  - **Bounded latency**: the Slack call runs inside the user's submit request, so the client is created with `retryConfig: { retries: 0 }` — the `@slack/web-api` default policy (ten retries over ~30 minutes on 5xx/429) would hang the submission during a Slack outage. One attempt, then give up; the helper never throws either way.
  - Message: emoji + company name, full suggestion text, submitter (respects the anonymous checkbox), page path, attachment.
  - **No truncation**: long suggestions are split across multiple section blocks (~2,900 chars each, word-boundary aware, up to 12 blocks ≈ 35k chars) instead of hitting Slack's 3,000-char section limit and failing silently; beyond the cap it appends a "read the full text in Carbon" note. Hard cuts back up so they never land inside an escaped entity.
  - User-controlled text is mrkdwn-escaped (no link injection into the channel).
  - Attachments live in the private bucket, so the message carries a 7-day signed URL.
- **i18n**: "Send to Carbon" translated across all 12 locales (both catalogs).

## Verification

- Typecheck green on `@carbon/env`, `@carbon/lib`, `erp`, `mes`; biome clean; `linguito check` clean (0 missing across all catalogs).
- Block-splitting logic verified with a runtime test: lossless reassembly, word-boundary breaks, hard-cut fallback, ≤2,900 chars per chunk.
- Slack client no-ops on localhost by design, so end-to-end delivery is verifiable only in production — the mechanism is identical to the already-working feedback widget.

## Notes for reviewers

- `packages/lib/AGENTS.md` says routes must dispatch Slack via `trigger("send-slack", ...)` — that path resolves the *tenant's* workspace token, which is exactly wrong here; the direct `getSlackClient()` call is deliberate and matches the pre-existing feedback routes. Follow-up candidate: document this exception in the AGENTS.md.

## Summary by CodeRabbit

* **New Features**
  * Added a **“Send to Carbon”** checkbox to suggestion submissions (shown only when enabled), alongside **“Submit anonymously”**.
  * When enabled, submissions are forwarded to Carbon via Slack with sanitized identity and attachment details; submission options reset after successful submit.

* **Localization**
  * Added **“Send to Carbon”** translations across supported languages (including German, Spanish, French, Hindi, Italian, Japanese, Korean, Polish, Portuguese, Russian, Turkish, and Chinese).


barbinbrad:
The suggestion box on the MES should have this too

Screenshot 2026-07-18 at 12 00 21 PM

barbinbrad:
For self-hosters, they shouldn't see this option. I think we should only show this if we have a SLACK_BOT_TOKEN enviornment variable set.

barbinbrad (changes requested):
Show checkbox in MES and only show checkboxes if SLACK_BOT_TOKEN is set

barbinbrad (approved):

PR #1173, opened 2026-07-20
fix: unify favicons across apps via shared @carbon/utils faviconLinks

aashu0148:
What does this PR do?

- Makes the favicon consistent across all apps (ERP, MES, Academy, Starter, Docs). Before, each surface could show a different mark because several competing icon files were shipped and the browser picked between them inconsistently.
- Uses two theme-aware brand marks (light mark on light theme, dark mark on dark theme) and fixes the old mapping, which was backwards.
- Deletes the extra icon files; Apple Touch and PWA install icons reuse the same brand mark.
- Centralizes the icon <link> tags in one shared faviconLinks (@carbon/utils) so all five apps stay in sync instead of drifting from copy-pasted markup.

> PDF response with no HTML head to attach an icon to. Theme switching only applies to pages with an HTML head, not the manifest/Apple Touch icons.

Images:
image
image

## Summary by CodeRabbit

- **New Features**
  - Added consistent, theme-aware favicon, touch-icon, and web manifest support across Academy, ERP, MES, Starter, and documentation sites.
  - Updated web app manifests to use the Carbon mark icon.
- **Bug Fixes**
  - Improved favicon and manifest loading consistency across supported applications.


packages/utils/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "exports": {
     ".": "./src/index.ts",
+    "./favicon": "./src/favicon.ts",
```
barbinbrad: we don't need both barrel style and path style exports. one or the other is preferred.

barbinbrad: i'd say let's remove the barrel file export and keep this one.

aashu0148:
@barbinbrad resolved but i kept the barrel export and removed the external path mention -- keeping it consistent across the app

barbinbrad (approved):

PR #1186, opened 2026-07-21
fix(audit): composite-key event pairing, tiptap diffs, FK snapshots

naveenkash:
## Summary

Audit-log correctness + readability overhaul, in four parts:

### 1. Composite-key event pairing
`dispatch_event_batch` now pairs UPDATE transition rows on the table's full primary key (or, when no PK exists, its smallest **valid, non-partial, non-expression** unique index — key columns only). This fixes the old/new cross-join on composite-identity tables like `itemPlanning` that emitted spurious audit diffs during bulk updates.

### 2. Schema-driven FK snapshots
FK changes are frozen into the diff as human-readable display values ("Supplier: Acme → Bolton") that survive later renames/deletes. Previously this required a hand-written per-column allowlist (15 columns covered out of ~280 business FKs on audited tables); now:

- **`get_foreign_key_map`** (new SQL function) discovers the FK topology of audited tables from `pg_constraint` at runtime — handles composite FKs `(col, companyId) → (id, companyId)`, excludes non-`id` references, and reports whether the target is tenant-scoped (`user` is not).
- **`fkDisplayRegistry`** (audit.config.ts) is the only human-maintained piece: display columns per FK *target* table (`supplier: ["name"]`, `user: ["fullName"]`, …) — one line per table, reused by every FK pointing at it. Typechecked against the generated DB types.
- Per-column `snapshotFields` remain as overrides and win over the registry. Unregistered targets degrade gracefully to the raw id.

### 3. Diff noise suppression
A form's first save writes `""` / `{}` / `[]` over DB nulls for every untouched field, which logged junk lines like `Custom Fields: null → {}`. The diff engine now skips **empty↔empty** transitions entirely (null/undefined/`""`/`{}`/`[]` are equivalent "empty"); real set/clear transitions still log, and `0`/`false` count as real values. Diff computation moved to a pure, unit-tested `events/diff.ts` module. Tiptap rich-text columns are diffed as extracted plain text (atomic), with malformed docs degrading to `""` instead of poisoning the Inngest batch.

### 4. UI polish
Both audit renderers (record drawer + Settings → Audit Logs table) show a muted italic **Empty** pill instead of the literal `null` — localized in all 13 languages. A blank snapshot display value is treated as absent, and the empty-value predicate is shared via `~/components/AuditLog`.

**Migrations:** `20260717143448_fix-event-update-row-pairing.sql`, `20260724130930_audit-fk-map.sql`
**Tests:** 21 new unit tests (`fk-snapshots.test.ts`, `diff.test.ts`) — 51 total green in `@carbon/jobs`.

### Before — raw ids, literal `null`, empty→empty noise
`Custom Fields: null → {}` and blank `Tracking Number: null →` lines on every first save; FK ids like `pterm_C5Ahjy…` shown raw:

Screenshot 2026-07-27 at 1 26 12 PM

### After — first-time sets
Empty sides render as a muted *Empty* pill; FK targets resolve to names (Net 15, Digi Key HQ, Air); no noise lines:
Screenshot 2026-07-27 at 1 15 29 PM

### After — value-to-value FK change
Both sides of an FK change show frozen display names (`Net 15 → Net 30`), and clearing a field reads naturally (`123 → Empty`):
Screenshot 2026-07-27 at 1 25 01 PM

 

## Summary by CodeRabbit

* **New Features**
  * Audit history now uses human-readable display snapshots for related foreign-key fields.
  * Audit log UI renders a localized **“Empty”** pill for empty diff values instead of showing raw empties.
* **Bug Fixes**
  * Corrected audit event old/new row pairing for tables with composite identities.
  * Improved diff generation by suppressing noisy empty-to-empty transitions and treating rich-text changes as atomic.
* **Documentation**
  * Updated audit log docs to clarify FK display snapshot behavior and precedence.
* **Tests**
  * Added unit tests covering empty vs real-value diff transitions.


sidwebworks (approved):
LGTM - just fix the conflicts and I think we're good

PR #1219, opened 2026-07-24
fix(auth): show company picker on passkey sign-in for multi-company users

sidgaikwad:
## Problem

Signing in with a passkey dropped multi-company users straight into an arbitrary first company instead of showing the company picker.

The authenticated layout (`x+/_layout.tsx`) uses the `companyId` cookie as the "has chosen a company this session" marker — if it's present, the picker is skipped. The passkey verify route set that cookie **unconditionally** after authentication, so multi-company users never saw the picker.

The magic-link callback (`_public+/callback.tsx`) already gets this right: it sets the cookie only for single-company users, deliberately leaving it unset for multi-company users so the layout routes them to `/select-company`.

## Fix

Mirror the callback contract in `api+/passkey.authenticate.verify.ts`: only set the `companyId` cookie when the user belongs to a single employee company. Multi-company users get the cookie left unset and are routed to the picker by `x+/_layout`.

- **Single-company users** — unchanged: cookie set, straight into the app.
- **Multi-company users** — cookie left unset, redirected to `/select-company`, exactly like the email/magic-link flow.

The session still carries a valid `companyId` (from `signInWithPasskey`) for the JWT; only the client-side "chosen" cookie is now conditional.

## Verification

- `pnpm exec turbo run typecheck --filter=erp` — the changed file is clean (remaining errors are pre-existing, unrelated dependency-resolution issues in `BoMExplorer` / `packages/viewer`).

Not yet browser-verified end-to-end (needs a live dev stack with a multi-company passkey user).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved passkey sign-in by looking up the signing-in employee’s associated companies before redirecting.
  * If the employee is linked to a single company, the app now automatically selects it; otherwise it prompts for company selection.
  * The authentication session cookie is always included in the final redirect response.
  * The company cookie is set only when applicable (single-company case).
  * If company lookup fails, sign-in stops with an error instead of completing cookie finalization.
  * Safe redirect behavior is preserved.


naveenkash (approved):
LGTM

Issue #1047, opened 2026-07-04
Accounting: record integrity & audit hardening — payment/memo immutability, always-on audit, balance constraint, JE export, book seed


barbinbrad:
## Code-verified status — 2026-09-09

**Verified on main:** Audit-log UPDATE/DELETE protection exists on main, including a controlled archival deletion path. Existing application/RLS guards cover some posted records, but the specified comprehensive database control layer is absent.

**Completed #1599 contribution / scope boundary:** PR #1599 adds FOR UPDATE payment/memo posting and exact-reversal/refund safeguards plus trusted author injection for settlement APIs. These improve real call paths without supplying trigger-level immutability, synchronous audit or COMMIT-time balance enforcement.

**Remaining:** Add service-role-proof payment/memo/settlement immutability, a deferred Posted-journal balance constraint using account-class signs, mandatory synchronous accounting audit coverage, protected seven-year retention, canonical uncapped streaming JE export with opening TB + population = closing TB proof, book seed/default-fill, and audited Repair entries. Audit capture still runs asynchronously and skips companies with auditLogEnabled=false; the archive threshold of 30 days is not evidence of total-history deletion, but required protected seven-year retention was not found.

**Source:** [append-only audit trigger](https://github.com/crbnos/carbon/blob/78239f8bf8a1f9c83efee1e7ce35cc35c6319009/packages/database/supabase/migrations/20260818014100_audit-log-append-only.sql#L12); [optional asynchronous audit capture](https://github.com/crbnos/carbon/blob/78239f8bf8a1f9c83efee1e7ce35cc35c6319009/packages/jobs/src/inngest/functions/events/audit.ts#L90); [archive upload uses mutable storage](https://github.com/crbnos/carbon/blob/78239f8bf8a1f9c83efee1e7ce35cc35c6319009/packages/jobs/src/inngest/functions/scheduled/audit-archive.ts#L70); [payment UPDATE RLS lacks posted-column trigger](https://github.com/crbnos/carbon/blob/78239f8bf8a1f9c83efee1e7ce35cc35c6319009/packages/database/supabase/migrations/20260630093809_ar-ap-payments.sql#L271); [branch transactional payment lifecycle](https://github.com/crbnos/carbon/blob/369d1ddfdbb890562937cb82dcb72d1f20723254/packages/database/supabase/functions/post-payment/post-payment-transaction.ts#L94).

**Other PRs:** No separate matching open implementation PR identified in the repository’s current open-PR listing.

[...]

Issue #1038, opened 2026-07-04
Accounting: gapless journal numbering + legal invoice series


barbinbrad:
## Code-verified status — 2026-09-09

**Verified on main:** The shared edge allocator and SQL RPC now increment counters atomically, closing the old SELECT-then-UPDATE duplicate race. This is already on main.

**Completed #1599 contribution / scope boundary:** PR #1599 does not deliver posting-time numbering or legal-series/cutover controls.

**Remaining:** Invoice creation still calls get_next_sequence before inserting the draft. Complete the coordinated six-sequence move to posting-time allocation, Draft placeholders, rollback/no-gap proofs, used-counter format/rewind/delete protection, gaplessFrom cutover and legalSeries allocation/selection, plus export backdating flags. The existing open numbering PRs are unmerged; #1227 overlaps/succeeds the #1182 foundation and must be reconciled, not both blindly landed.

**Source:** [atomic edge allocator](https://github.com/crbnos/carbon/blob/78239f8bf8a1f9c83efee1e7ce35cc35c6319009/packages/database/supabase/functions/shared/get-next-sequence.ts#L28); [atomic SQL allocator](https://github.com/crbnos/carbon/blob/78239f8bf8a1f9c83efee1e7ce35cc35c6319009/packages/database/supabase/migrations/20260730185058_fix-get-next-sequence-race.sql#L5); [draft purchase invoice allocates before insert](https://github.com/crbnos/carbon/blob/369d1ddfdbb890562937cb82dcb72d1f20723254/apps/erp/app/modules/invoicing/invoicing.service.ts#L651).

**Other PRs:** [#1182](https://github.com/crbnos/carbon/pull/1182) (OPEN, draft) [#1227](https://github.com/crbnos/carbon/pull/1227) (OPEN, unmerged)

**Evidence boundary:** Static source inspection against main `78239f8bf8` and PR #1599 HEAD `369d1ddfdb`. No new runtime tests or database mutations were performed for this issue-status audit; open PR metadata/file inventories are cross-references, not completion evidence.

Tracking basis: per Brad’s instruction, treat #1599 as merged. Completion credits use the verified implementation at `369d1ddfdb`; source links to `78239f8bf8` describe the pre-merge foundation. The outstanding work listed above remains separate. Program overview: #1060.

---

Part of the **Accounting Readiness program** — meta spec: `.ai/specs/2026-07-04-accounting-implementation-meta.md` (PR #1013 until merged). Readiness finding SD-2.

[...]

PR #1227, opened 2026-07-24
feat(accounting): gapless numbering & legal series — substrate + service + UI (#1038)

carbon-agent:
## Gapless Numbering & Legal Series (#1038)

Tracking spec: `.ai/specs/2026-07-04-gapless-numbering-legal-series.md`
Readiness finding **SD-2**. Supersedes the deferred foundation draft #1182 — this
branch rebases that foundation onto current `main` and adds the safe, verifiable
next layer.

### What this delivers

**DB substrate (foundation migration `20260721162233`, rebased onto main)**
- `sequence` hardening: `isLegalSequence` / `firstUsedAt` / `gaplessFrom` columns
  + `sequenceImmutabilityCheck` trigger (**every role**, service role included) —
  freezes prefix/suffix/size/step/table and blocks rewinding `next` or deleting a
  used legal sequence. Closes the rewind→duplicate and format-fork holes.
- Atomic allocators `get_next_sequence_atomic` / `get_next_legal_series_number` —
  single `UPDATE … RETURNING`; the row lock is the serialization point, so
  concurrent posts can't duplicate and a rolled-back post leaves no gap. The
  shared edge helper `getNextSequence` uses the same shape, so every in-transaction
  poster is already race-free.
- `legalSeries` table (company × country × document type) + RLS + immutability
  trigger; nullable `legalSeriesId`/`legalNumber` on `salesInvoice`/`memo`.
- Draft number columns made nullable behind status CHECKs; per-company
  `gaplessFrom` forward-only cutover stamp. Migration is idempotent
  (`IF NOT EXISTS` / `DROP … IF EXISTS`).

**New in this branch**
- **Allocator REVOKE hardening** (migration `20260724161500`): REVOKE EXECUTE on
  both atomic allocators from `PUBLIC`/`anon`/`authenticated` — a `SECURITY
  DEFINER` function defaults to PUBLIC, which would expose them as PostgREST RPCs
  and let any client burn an accounting number in a standalone transaction (the
  SD-2 gap source). They now only run inside posting transactions. Verified no
  app/edge caller reaches them via PostgREST.
- **`legalSeries` service** — `getLegalSeries` / `getLegalSeriesById` /
  `getLegalSeriesList` / `upsertLegalSeries` / `deleteLegalSeries` (cast-based
  until the table lands in generated types).
- **Legal Series settings UI** — accounting-settings CRUD (list/new/edit/delete +
  form + table + nav), mirroring Payment Terms. Used series show frozen format
  fields (mirrors the DB trigger).
- **Sequence format lockdown UI** — sequences form disables prefix/suffix/size/step
  once `isLegalSequence && firstUsedAt`, so a rejected edit is a disabled field
  instead of a 500 from the trigger. `next` stays editable.
- **Draft placeholder primitive** — `getDocumentReadableId` /
  `getDraftDocumentPlaceholder` (`Draft-{id6}`) in `@carbon/utils`, unit-tested.
- i18n filled for all 12 non-English locales.

### Acceptance criteria

| # | Criterion | Status |
|---|-----------|--------|
| 3 | Per-company serialization, zero cross-company contention | ✅ atomic `UPDATE` (row-lock scoped to company × type) |
| 6 | `sequence` rows immutable after first use | ✅ trigger (all roles) + lockdown UI |
| 2 | Sequence edits rejected (+ logged) | ✅ trigger rejects; UI mirrors it. *Logged* = #1047 audit coverage |
| 7 | Migration idempotent | ✅ `IF NOT EXISTS` / `DROP … IF EXISTS` |
| — | RPC hardening | ✅ atomic allocators REVOKEd from PostgREST |
| 4 | All six sequences supported | ✅ allocator + `isLegalSequence` cover all six; `legalSeries` substrate for customer-facing docs |
| 1 | No gap on failure | ⚙️ atomic helper already gapless for in-trx posters; fully closed for manual docs once draft→post allocation moves (wave below) |
| 5 | Draft `Draft-{id6}` placeholder | ⚙️ tested primitive shipped; render-site wiring lands with the wave (numbers only go null then) |

### Remaining — the posting-time wave (Decision 15, DB-gated)

The spec calls this a "coordinated PR wave." Enabling the `get_next_sequence`
RAISE guard for the six accounting sequences requires first moving every
draft-creation allocator to posting time, or those callers throw. Full verified
call-site map (10 draft/burner sites + 6 posting paths, incl. the `postJournalEntry`
Kysely refactor) is in **`.ai/runs/1038/status.md`**. Kept separate because it
rewrites the critical accounting posting path and must be verified against a live
DB before merge.

### Verification

- ✅ `@carbon/utils` test (73 pass, incl. 5 new placeholder cases) + typecheck
- ✅ `erp` + `@carbon/database` typecheck (`tsgo --noEmit`)
- ✅ Biome, i18n `linguito check` clean
- ⚠️ **No Postgres/stack in this environment** — the two migrations were not
  applied and `generate:types` was not run. All DB-level behavior (SQL, triggers,
  CHECKs, the REVOKE, and every concurrency/rollback acceptance test) is unproven
  at runtime and must be verified against a live DB before merge. The service
  layer intentionally uses casts until `legalSeries` is in generated types.

## Summary by CodeRabbit

* **New Features**
  * Added Accounting **Legal Series** management (list/search, create, edit, delete) with drawer/table workflows and format-lock behavior after first use.
  * Introduced the statutory **gapless numbering foundation** with legal-series sequence settings and format locking in sequence UI.
* **Bug Fixes**
  * Updated numbering allocation to be atomic to reduce duplicate-number risk under concurrency.
  * Hardened protected sequence allocator access to prevent unintended exposure.
* **Localization**
  * Added/updated Legal Series and numbering-lock UI translations across supported languages.
* **Tests**
  * Added tests for draft readable-id placeholder helper behavior.


carbon-agent:
## Addressed review feedback (commit ddeaab8da)

All 12 actionable comments resolved:

| # | File | Fix |
|---|------|-----|
| 1 | `accounting.service.ts` | `getLegalSeries` search now escapes backslash/quote and interpolates as a **double-quoted PostgREST value**, so reserved chars (`,` `(` `)` `.` `:` `*`) can't split the `.or()` filter. |
| 2 | `LegalSeries/LegalSeriesForm.tsx` | Document-type labels moved inside the component and run through Lingui `` t`` `` (`Sales Invoice` / `Credit Memo` already exist in every catalog). |
| 3 | `mcp+/lib/tool-metadata.json` | **Root cause fixed in `scripts/generate-mcp.ts`**, not hand-edited: the `>` of an arrow function (`errorMap: () => ({…})`) was miscounted as a closing bracket, truncating validator parsing after `documentType`; a `z.preprocess()` around a non-inline enum also recursed forever. Regenerating restores the full `accounting_upsertLegalSeries` schema (`code`, `name`, `prefix`, `size`, `validFrom`, …) and corrects many other previously-truncated tool schemas. |
| 4 | `legal-series.$legalSeriesId.tsx` | Loader `throw notFound()` on missing/errored series instead of silently rendering the create form. |
| 5 | `legal-series.tsx` | Loader surfaces `getLegalSeries` errors instead of falling back to an empty list. |
| 6 | migration `20260721162233` | Deprecated `has_role('employee', …)` replaced with `get_companies_with_employee_role()` in both atomic allocators (matches this migration's own RLS). |
| 7 | migration `20260721162233` | The five new `*_posted_requires_number` CHECK constraints split into `ADD CONSTRAINT … NOT VALID` + `VALIDATE CONSTRAINT`, avoiding a blocking full-table scan on deploy. |
| 8 | `de/erp.po` | "Legal Series" → **"Belegnummernkreis"** across all 5 strings (grammar adjusted: der/den/Neuer). |
| 9 | `ko/erp.po` | Natural delete-confirmation wording. |
| 10 | `ko/erp.po` | "Registration Ref" → **등록 참조**. |
| 11 | `@carbon/utils/accounting.ts` | `getDocumentReadableId` falls back to the draft placeholder for empty/whitespace-only numbers; added empty + whitespace test coverage. |

**Verification:** `typecheck` (erp + @carbon/utils) ✅ · `@carbon/utils` tests 74/74 ✅ · biome lint ✅ · `turbo run build --filter=erp` ✅ (exit 0).

carbon-agent:
## Review feedback addressed

The 12 original review comments (posted 17:42Z) were already resolved in `ddeaab8`. CodeRabbit's re-review then flagged one **remaining** item on the `scripts/generate-mcp.ts` fix (comment on `generate-mcp.ts` L156), which this push (`dbede59b`) closes:

**`generate-mcp.ts` — token-aware delimiter stack.** The prior fix only special-cased the arrow `=>`. A *standalone* comparison operator inside an arrow function — e.g. a zod `.refine((v) => v > 0)` — still decremented the bracket-depth counter, so the next comma/colon could be misread as top-level and corrupt validator field-splitting (and `findTopLevelColon` had the same flaw). `splitAtTopLevel` and `findTopLevelColon` now share an explicit delimiter stack:
- `<` is pushed **only** when it opens a generic (it abuts an identifier or a preceding `>`, unlike a spaced comparison `a < b`);
- `>` pops **only** a matching `<`;
- parens/braces/brackets pop only their own opener — arrow `>` and comparison `<`/`>` are left untouched.

**Verification**
- Verified the two reviewer failure cases (`v > 0`, `v < 10` inside `.refine(...)`) plus a ternary-with-comparison default now split correctly; generics (`Record<K, V>`, `z.infer<typeof x>`) and arrow `errorMap` still track correctly. All pass; the old `=>`-only guard failed the comparison cases.
- Regenerating `tool-metadata.json` produces **byte-identical** output on the current codebase (no schema regression) — the fix hardens the generator against this bug class going forward.

Note on regression tests: the reviewer asked for tests on both helpers, but `scripts/` has no vitest project (standalone `tsx` scripts, functions not exported, not covered by `turbo run test`), so a test file there would not run in CI. Correctness was instead proven by the case-based verification above and the identical-output regeneration.

carbon-agent:
## Addressed remaining review feedback (commit 228a492b9)

The 12 original comments (17:42Z) and the first `generate-mcp.ts` follow-up were already resolved in `ddeaab8` / `dbede59`. CodeRabbit's latest re-review (19:39Z) flagged **one residual variant** on `scripts/generate-mcp.ts`, which this push closes:

**`isGenericOpen` — token-aware generic detection instead of character adjacency.** The previous version treated any `<` abutting an identifier/`>` as a generic open, so compact comparisons — `v<10`, `v<=10`, `x=>y<z` — pushed a `<` that never popped, unbalancing the delimiter stack and corrupting later comma/colon splitting.

A generic `<` now must be confirmed by a **balanced forward `>` scan** (`hasGenericClose`) that ignores arrow `=>` and tracks nested `()`/`[]`/`{}`, plus a cheap `<=` guard. A comparison has no matching `>`, so it is left as an ordinary token.

**Verification**
- Case checks (standalone `tsx`): all three reviewer cases (`v<10`, `v<=10`, `x=>y<z`) now split/colon-detect correctly, and generics still track — `Record<K, V>`, `SupabaseClient<Database>`, `z.infer<typeof x>`, function-type generics `Array<() => string>`, and nested `Record<K, Map<A, B>>`. The old adjacency check failed the comparison cases.
- Regenerating `tool-metadata.json` is **byte-identical** (only the `generated` timestamp differs, reverted) — no schema regression; the fix hardens the generator against this bug class.

**On regression tests:** `scripts/` is a standalone `tsx` directory with no vitest project (helpers unexported, not covered by `turbo run test`, ignored by biome), so a co-located test file would not run in CI. Correctness is instead proven by the case-based verification above and the identical-output regeneration. The change is isolated to this build-time generator and is not imported by any app, so `@carbon/erp` compilation is unaffected.

PR #1235, opened 2026-07-24
fix(erp): match multi-word and untrimmed list searches

### LINKED ISSUE #1234: Item list search misses multi-word queries ("M8 Washer" doesn't find "Washer, Flat, M8")
### Issue Summary

On the Inventory and Items (Parts/Materials/Tools/Consumables/Services) list screens, the search box only matches when the typed text appears as one contiguous substring of the record. A multi-word query whose words appear in a different order than the record's name never matches, and a trailing or leading space in the search box breaks matching entirely.

### Steps to Reproduce

1. Go to Inventory (or Items > Parts).
2. Find or create an item named e.g. `Washer, Flat, M8`.
3. Type `M8 Washer` into the search box.
4. Also try `M8 Washer ` (trailing space) and `M8` (single word, for comparison).

### Actual Results

- `M8 Washer` returns nothing, even though every word in the query appears in the item's name.
- `M8 Washer ` (trailing space) also returns nothing.
- `M8` alone works fine.

### Expected Results

- `M8 Washer` and `M8 Washer ` should both return `Washer, Flat, M8` — every word the user typed is present in the record, just in a different order. Carbon's own global (Cmd+K) search already behaves this way: it trims whitespace and matches on all words in any order (see `packages/database/supabase/migrations/20260419140000_search-trigram-substring.sql`).

### Technical details

- Server-side query construction, not browser-specific. Each list service (`apps/erp/app/modules/inventory/inventory.service.ts`, `apps/erp/app/modules/items/items.service.ts`) interpolates the whole `search` param into a single `%...%` PostgREST `ilike` pattern, so it can only match a contiguous substring, and nothing trims the value first.

### Evidence

- Confirmed by reading the query construction directly — every affected function does `query.or(\`name.ilike.%${args.search}%,...\`)` with the raw, untrimmed string.

I have a small fix ready and will open a PR shortly.

layne-sygnet:
## What does this PR do?

- Fixes #1234

Item list search (Inventory, Parts, Materials, Tools, Consumables, Services) only matched when the typed query appeared as one contiguous substring, so searching `M8 Washer` never found a part named `Washer, Flat, M8`, and a trailing or leading space broke matching entirely.

This adds a small `setSearchFilter` helper (`apps/erp/app/utils/query.ts`) that splits the search term into whitespace-separated tokens and ANDs one `ilike` OR-group per token, so every token must match somewhere across the searched columns, in any order. This mirrors the tokenization Carbon's own global (Cmd+K) search already does — `search_company_index()` in `packages/database/supabase/migrations/20260419140000_search-trigram-substring.sql` trims, collapses whitespace, and AND-joins tokens into a tsquery. This PR brings the list-screen search path in line with that existing behavior rather than introducing a new one.

Also trims the search box value before it reaches the URL (`SearchFilter.tsx` in both `apps/erp` and `apps/mes`), so leading/trailing spaces never affect matching, on every list screen that uses the shared search box.

**Single-word searches are unchanged** — the new code produces the exact same query PostgREST sees today when the search term is one word (covered by a unit test).

**Scope:** this PR touches 7 call sites — Inventory quantities (×2) and Parts/Consumables/Materials/Tools/Services. There are roughly 75 other `.ilike` search call sites elsewhere in the app (sales, purchasing, quality, receipts, shipments, etc.) with the same underlying pattern. Happy to follow up with those separately, or to narrow/widen this PR's scope if you'd prefer a different cut.

**On query cost:** a 2-word search now runs 2 `ilike` OR-groups instead of 1, both against the same unindexed columns as before. Since AND-ing tokens is strictly more selective than the old single-substring match, this shouldn't be slower in practice, but I don't have production-scale data to confirm — flagging it in case it's worth a second look.

## Visual Demo

No screen recording available in this environment. Reproduction steps and expected before/after behavior are in "How should this be tested?" below — happy to add a screen recording if that would help review.

## Mandatory Tasks

- [x] I confirm automated tests are in place that prove my fix is effective

## How should this be tested?

- Unit test: `pnpm --filter erp exec vitest run app/utils/query.test.ts` — covers `getSearchTokens` and `setSearchFilter` directly, including the exact clause strings emitted for multi-word, whitespace-only, and single-word inputs.
- **Note:** `apps/erp` doesn't currently have a `test` script wired into the root `pnpm test` / CI (its 13 existing test files aren't run there either), so this test runs via the command above but not automatically in CI. Flagging that gap rather than leaving it silent — happy to help wire it up if useful, as a separate PR.
- Manual repro: on Inventory or Items > Parts, find/create an item named e.g. `Washer, Flat, M8`. Search `M8 Washer` — it now appears. Search `M8 Washer ` (trailing space) — same result. Search a single word like `M8` — behavior is unchanged from before this PR.
- Locally passing: `pnpm exec biome check`, `pnpm exec turbo run typecheck --filter=erp --filter=mes`, `pnpm run lingui:check` (no `.po` diff — no user-facing strings changed), `pnpm test` (21/21 packages).

## Checklist

*(all applicable items followed; nothing to flag)*

## Summary by CodeRabbit

* **New Features**
  * Improved search across inventory and item listings, including multi-word searches and relevant item identifiers.
  * Search handling now sanitizes special characters for safer, more reliable results.

* **Bug Fixes**
  * Leading and trailing spaces are removed from search terms before updating the URL.
  * Empty or whitespace-only searches no longer apply unnecessary filters.


barbinbrad (approved):

barbinbrad:
Thanks for this fix!

PR #1246, opened 2026-07-27
fix(erp): job BoP save scroll + step/tool forms match the item editor

sidgaikwad:
## Problem

On the job details page, saving an **invalid** operation or step jumps the view up to the Bill of Material instead of showing the field that failed validation. The step form also had two separate defects. The item/part editors were reported as working correctly, and they are — the difference turned out to be the key to the whole thing.

## Before:
image

- in this after adding first step the modal then collapses and just show the add step button only
- and on invalid save it takes us to BOM

## After:

https://github.com/user-attachments/assets/ab32cc3e-76fe-472c-9974-17b8b00d6f68

## Root cause (confirmed)

**1. The scroll jump lives in `@carbon/form`, not in the job page.**

When validation fails, `focusFirstInvalidInput` scrolls the first invalid field into view. It has two branches — scroll a specific container, or fall back to native `scrollIntoView`. The container branch did:

```ts
container.scrollTo({ top: element.offsetTop })
```

`offsetTop` is measured from the element's nearest **positioned** ancestor, not from the container. `SortableList` renders every BoM/BoP row as a framer-motion `Reorder.Item` with an inline `position: relative` (`SortableList.tsx:105,109`), so for a field inside an operation card `offsetTop` is the small intra-card offset — tens of pixels. The panel scrolled to ~40px from the top of its content, which is exactly where the Bill of Material sits.

Why only the job page reproduces it: the container branch is selected by a Tailwind class heuristic, `.overflow-hidden > .h-full`. `x+/job+/$jobId.details.tsx` is the one page whose scroller is `h-full` inside an `overflow-hidden` parent. Part, tool, service and quote all use `h-[calc(100dvh-99px)]`, never match the selector, and fall through to native `scrollIntoView` — which is why they behave correctly.

Fixed by computing the offset from viewport rects and centring the field, matching the `block: "center"` fallback. I fixed the arithmetic rather than removing `h-full` from the job route, since that would only hide a broken helper until the next page happened to match the selector.

**2. List steps failed validation silently.**

`operationStepValidator`'s List refine attached its error to `path: ["listOptions"]`, but the field is named `listValues`. Field errors are looked up by field name (`useFieldError`), so saving a List-type step with empty options produced no message and no focus — you click Save Step and nothing happens. This one is in the shared validator, so it affected the item and quote editors too.

**3. Step form reset to the wrong type.**

The job and quote step forms reset the type select to `"Value"` after a save; the working item editor resets to `"Task"`. The quote form additionally reset the tiptap description to `[]` instead of `{}`, where a document object is expected.

## Quote BoM/BoP

Checked, since it was asked about. Quote does **not** have the scroll bug — its content wrapper isn't `h-full`, so it never enters the broken branch. It did share both step-form defects, and those are fixed here.

## Changes

| File | Change |
|---|---|
| `packages/form/src/ValidatedForm.tsx` | Scroll the container from viewport rects instead of `offsetTop` |
| `packages/form/src/internal/scrollIntoView.ts` | New — the offset arithmetic, with the offsetParent trap documented |
| `apps/erp/app/modules/shared/shared.models.ts` | List-options error targets `listValues`, not `listOptions` |
| `apps/erp/app/modules/production/ui/Jobs/JobBillOfProcess.tsx` | Reset step type to `Task` |
| `apps/erp/app/modules/sales/ui/Quotes/QuoteBillOfProcess.tsx` | Reset step type to `Task`, description to `{}` |
| `.ai/lessons.md` | Lesson on `offsetTop` vs container scroll offsets |

## Verification

`biome check` clean · `typecheck` passes for `@carbon/form` and `erp` · `@carbon/form` tests 10/10.

The corrected scroll maths was developed against a temporary unit test that reproduced the job-page geometry (pre-fix it produced a scroll target of 40 where 1621 was needed); the test was dropped before merge at the author's request, so **this change ships without automated coverage for the scroll offset** — worth a manual pass on a job with a BoP operation.

## Not fixed here

`apps/erp/app/modules/quality/quality.models.ts:633` has the identical `listOptions`/`listValues` path bug for inspection attributes. Left alone as out of scope — same one-word fix if wanted.

## Follow-up: the three copies were forks, and had drifted

The reporter asked why a job/quote fix needs so many files, given item/part works. It doesn't have to — `BillOfProcess`/`BillOfMaterial` are copy-paste forks across item/job/quote (~13k lines, 61–79% identical), and fixes keep landing in one copy but not the others. The `setType` reset bug above is one example; the card-styling fix took two commits and *still* never reached job's parameters list.

Rather than a consolidation refactor (deferred), this brings the job and quote copies back in line with the working **item** editor:

- **Job's add-step form** was hidden behind an `Add Step` disclosure button (added to the job copy only, Sep 2025); now rendered inline like item and quote. This is the "no inline step form on jobs" the reporter pointed at.
- **ToolsForm** (job + quote): restored the `{!isDisabled && …}` guard and `autoFocus={tools.length === 0}` that only item had.
- **Rounded first/last-card styling** applied to job's parameters list (item and quote had it; job never got it).
- **Five `<Trans>` wrappers** added to the "Please save the operation…" alerts (job steps/params, quote steps/params/tools) — same message ids as item, no new translations.

No behaviour change on saved operations: the add forms were already gated by a disabled Submit for unsaved operations, so this only hides the non-functional form and matches item.

A full consolidation of the leaf sub-forms (~2.5–3.5k lines removable) is a sensible next step but was left out here; `OperationForm` and the containers carry genuine domain divergence (job: production events/chat/procedure-sync/status gating; item: the configurator) and shouldn't be force-merged.

## Summary by CodeRabbit

* **New Features**
  * Improved step, parameter, and tool entry flows across production and sales forms.
  * Add controls now appear only when the related forms are enabled.
  * Unsaved-operation alerts support translated text consistently.

* **Bug Fixes**
  * New entries reset to the appropriate task type and clear their content correctly.
  * Tool field focus is more consistent when adding the first tool.
  * Validation errors now identify the correct list-options field.
  * Invalid fields are centered more reliably within scrollable form areas.


barbinbrad:
Please provide a video demo for this type of PR and fix merge conflicts

sidgaikwad:
solved the merge conflicts and also attached a video for it

barbinbrad:
> solved the merge conflicts and also attached a video for it

Thanks, looks good. Do we need this for the `BillOfProcess.tsx` component that lives on the item master?

sidgaikwad:
no item master does'nt have this bug

aashu0148:
resolved here: https://github.com/crbnos/carbon/pull/1334

PR #1260, opened 2026-07-28
feat(items): CSV bulk import for material properties

sidgaikwad:
## What & why

Material taxonomy (Substances, Shapes/Forms, Types, Finishes, Grades, Dimensions) could only be entered one row at a time through the config forms. This adds a **per-page "Bulk Import" CSV importer** to each of the six config screens, reusing the existing Parts-import machinery — so each gets a **downloadable template**, a column/enum-mapping wizard, and results modal for free.

Behaviour is **create-only, skip-duplicate** (as requested — never creates duplicates): a row matching an existing entry for the company **or** a global system row (`companyId IS NULL`) is reported as *skipped*; re-importing the same file is a no-op.

## Video
https://drive.google.com/file/d/1BLWkRUNaCDDTPQI_RuV4LuqxzV9EIAHt/view?usp=sharing

## How it works

The importer is registry-driven, so most of this is registration:

- **`apps/erp/app/modules/shared/imports.models.ts`** — registers the six taxonomy tables in `fieldMappings`, `importPermissions` (→ `parts`), and `importSchemas`. Child taxonomies reference their parent substance/shape **by name** via enum fetchers, resolved to ids by the mapping step (parents must pre-exist).
- **`packages/database/supabase/functions/import-csv/material-property-import.ts`** (new) — a config-driven importer with per-table dedup keyed on the DB unique constraints (`code` for substance/form; `(substance, name)` for finish/grade; `(form, name)` for dimension; both `(substance, form, code)` and `(substance, form, name)` for type), scoped to the company OR global rows. Wired into the `import-csv` edge function's table enum + switch.
- **Six taxonomy `<Table>` components** — surface the import button via the existing `importCSV` prop.

No DB migration (all six tables already exist); no new i18n strings.

## Verification

- `pnpm exec turbo run typecheck --filter=erp` — green.
- Biome clean; lingui extract produces zero `.po` diff (labels reuse existing strings).
- **End-to-end against the local stack (Substances):** uploaded a CSV with a duplicate of the global `ALUMINUM/AL` plus one new row → **1 inserted, 1 skipped** ("Already exists — skipped"); re-importing the same file → **0 inserted, 2 skipped** (idempotent). The inserted row was company-scoped (CUSTOM) with a DB-generated id.

## New Features

* Added create-only CSV import support for Material Substances, Shapes, Finishes, Grades, Types, and Dimensions.
* Added clearly labeled CSV import options to each corresponding material table.
* Imports validate required fields and parent relationships, skip existing or duplicate records, and report inserted, skipped, and failed rows.
* Material Dimensions imports support metric status values.

## Documentation

* Expanded CSV import guidance with supported tables, parent resolution, duplicate handling, and existing-record behavior.


barbinbrad:
With a PR like this, please provide a video demo in the description

sidgaikwad:
Added the video

aashu0148 (approved):

PR #1267, opened 2026-07-29
fix(import): resolve material Finish, Grade, and Dimensions on CSV import

sidgaikwad:
## Problem

Bulk-importing materials via CSV silently drops the **Finish**, **Grade**, and **Dimensions** columns. The mapping UI sends them as raw text under `finish` / `grade` / `dimensions` (see `fieldMappings.material` in `apps/erp/app/modules/shared/imports.models.ts`), but the `import-csv` edge function's `materialValidator` only reads `finishId` / `gradeId` / `dimensionId`. Zod strips the unknown text keys, so all three attributes vanish with no error on **both** the insert and the update path. Substance and Shape import fine because they're enum-mapped to real ids in the mapping UI.

## Why not enum-map them like Substance/Shape

`materialFinish` and `materialGrade` are scoped **per substance**, and `materialDimension` **per form** (unique on `(scope, name, companyId)`). The same name legally exists under multiple scopes (e.g. *Hot Rolled* under steel and stainless), so the flat CSV-value → option enum mapping can't target them correctly for mixed-substance imports.

## Video
https://drive.google.com/file/d/1ANLSaL3ci-4SmMWzH75zEto5m--UvEo9/view?usp=drive_link

## Fix

Resolve the raw text server-side in the edge function, per row, within the row's substance/form scope:

- Extend `materialValidator` to accept the `finish` / `grade` / `dimensions` text the UI actually sends (explicit `*Id` values still win when provided).
- New `resolveMaterialTaxonomyIds()` bulk-resolves the distinct `(scope, name)` pairs inside the import transaction: case-insensitive match against global (`companyId IS NULL`) and company rows, company rows winning a name clash.
- Unmatched names **create a company-scoped taxonomy row** — the same behavior as the creatable Finish/Grade/Dimension comboboxes on the material form. Creation relies on the DB id defaults (`id('finish')` / `id('grade')` / `id('mdim')`) and uses `ON CONFLICT DO NOTHING` + re-select so a concurrent import can't fail the batch.
- Wire the resolved ids into both the material insert and update payloads.
- Rows with no resolved substance (finish/grade) or form (dimensions) leave the attribute unset — the taxonomy tables require the scope (`materialSubstanceId` / `materialFormId` are NOT NULL).

Also scopes the material update statement by `companyId`: `material`'s PK is `(id, companyId)` and `material.id` is the item's per-company `readableId`, and Kysely bypasses RLS — matching on `id` alone could touch another tenant's row with the same part number. The sibling item-type-table update just above it already carried this guard.

Documented the resolution behavior in `.claude/rules/csv-import-system.md`.

## Validation

- `biome check` on the edge function: clean (same 11 pre-existing warnings as `main`, none in the new code).
- Compiled the new query shapes against kysely to verify the emitted SQL:
  - `select "id", "name", "companyId", "materialSubstanceId" from "materialFinish" where "materialSubstanceId" in ($1) and ("companyId" = $2 OR "companyId" IS NULL)`
  - `insert into "materialFinish" ("name", "companyId", "materialSubstanceId") values ($1, $2, $3) on conflict ("materialSubstanceId", "name", "companyId") do nothing`
- Verified the `ON CONFLICT` targets match the live unique constraints (`materialFinish_name_companyId_unique`, `materialGrade_name_companyId_unique`, `materialDimensions_name_companyId_unique`) — later migrations only recreated the `companyId` FKs, not the uniques.
- Kysely ignores `undefined` in `set()`/`values()`, so CSVs that don't map these columns leave existing values untouched (unchanged behavior).

## Summary by CodeRabbit

* **New Features**
  * CSV imports now support Finish, Grade, and Dimensions as text values, resolving them to the appropriate taxonomy.
  * Values are matched case-insensitively with company-specific entries prioritized; missing values are created automatically.

* **Bug Fixes**
  * Improved material attribute handling for updates and new imports, preventing scope mismatches.

* **Documentation**
  * Clarified text handling and fallback behavior when required substance or form values are missing.


sidgaikwad:
@aashu0148 added the video link in the description plz check

aashu0148 (approved):

PR #1275, opened 2026-07-30
fix(invoicing): recompute invoice Due Date on Issue Date / Payment Term edit

sidgaikwad:
## Problem

Editing a sales or purchase invoice's **Issue Date** never updated its **Due Date**, and switching the **Payment Term** didn't either. In both `sales-invoice+/update.tsx` and `purchase-invoice+/update.tsx`, the `dateIssued` handler looked up the payment term with the new *value* as the term id:

```ts
.from("paymentTerm").select("*").eq("id", value as string)  // value is the date string
```

That lookup always missed, so the recompute branch was skipped and the handler silently saved only the date (request still returns 200). `paymentTermId` changes fell straight through to the generic `[field]: value` update with no recompute at all.

Because **Due Date drives AR/AP aging and payment follow-ups**, invoices kept a stale due date after any Issue Date or Payment Term edit — potentially weeks off — with no visible error.

## video

https://github.com/user-attachments/assets/105a5182-fa6f-460b-979d-71b2b033d915

## Fix

- **Date Issued edit**: read the invoice's *current* `paymentTermId`, look up that term, and set `dateDue = dateIssued + daysDue`.
- **Payment Term edit**: new dedicated case — the value here *is* a real payment term id; read the invoice's *current* `dateIssued` and recompute `dateDue`.
- Both fall back to a plain field update when the invoice has no payment term / no issue date, or when the value is cleared.
- Removed `paymentTermId` from the generic fall-through group so the new case handles it.

Applied identically to both the sales-invoice and purchase-invoice handlers.

## Repro (before → after)

1. Open any invoice with a Net-30 term.
2. In the right Properties panel, change **Date Issued** (or the **Payment Term**).
3. **Before:** Due Date doesn't move. **After:** Due Date recomputes to issue date + term days.

## Verification

- `pnpm exec turbo run typecheck --filter=erp` — passes.
- Biome + lingui pre-commit gates green (no missing translations).

## New Features
* Invoice due dates now automatically recalculate when the issue date or payment terms change.
* Due-date calculations support net terms, month-end terms, and specific days of the month for purchase and sales invoices.

## Bug Fixes
* Invoice date fields now reliably reflect the latest saved values.
* Updates are restricted to the correct company’s invoices.
* Invalid or incomplete date and payment-term information is handled safely without producing incorrect due dates.


barbinbrad:
This looks good, but we have other types of payment terms besides Net. 

For instance there End of Month + and Day of Month

Screenshot 2026-07-31 at 9 28 41 PM

I think we should handle those here too.

barbinbrad (changes requested):
Please handle end of month and day of month calculations too.

sidgaikwad:
done with the requested changes

barbinbrad (approved):

PR #1293, opened 2026-07-31
refactor(edge-functions): centralize HTTP response construction into lib/response.ts

sidwebworks:
## What does this PR do?

Centralizes HTTP response construction for all 40 Supabase edge functions behind a single `lib/response.ts` (`corsPreflight` / `jsonResponse` / `errorResponse` / `isDataLayerError`), replacing ~179 hand-rolled `new Response(...)` calls and pinning one wire contract the apps already read (`{ message }` + non-2xx on failure). Along the way it fixes real bugs the drift had accumulated: ~20 `JSON.stringify(err)` sites that shipped `"{}"` now surface the real message, `embedding` responses gain their missing CORS headers, `create` gets a top-level try/catch so an invalid payload / permission failure returns a proper `400`/`500` with headers instead of a bare CORS error, and `export-company`'s non-Error `[object Object]` body is fixed. `errorResponse` structurally sanitizes data-layer errors (Postgres / Postgrest / node-pg / Zod — by class/shape, never message text) so raw SQL never reaches a toast, while every authored `throw new Error(...)` still surfaces. Net −503 lines; every status code, implicit-200, the 4 protected binary/string responses, and `mrp`'s nested try/catch are preserved byte-for-byte.

- No linked issue — internal refactor prompted by 13× copy-pasted error blocks in `create/index.ts`.

## How should this be tested?

These files are Deno and are not covered by repo CI (not in `tsconfig` include or `biome` includes), so gates are local. From `packages/database/supabase/functions/`:

- `deno test --no-lock` — 97 pass (8 existing suites + a new 18-case `lib/response.test.ts`, including a sanitizer test proving classification is structural, not text-matching).
- `deno check --no-lock <file>` — per-file type-error counts unchanged vs the pre-refactor baseline for every touched file (no regressions).
- Grep invariants: exactly 8 residual `new Response` (the protected binary/string allowlist), `corsPreflight` adopted in 38 files, `corsHeaders` import surviving in exactly 3.
- Runtime smoke with `crbn up`: `OPTIONS .../create` → `200 ok`; `POST .../create -d '{}'` → `400 {"message":"Invalid payload"}` with CORS headers.

No env vars or test data required. One known behavior left flagged with `// FIXME:` (not changed): `post-production-event` returns `{ success: false }` at HTTP 200 — a status-semantics decision for a follow-up.

## Checklist

- Followed the project style (double quotes, 2-space indent, trailing commas), commented only non-obvious "why", and introduced no new warnings.

## Summary by CodeRabbit

* **Enhancements**
  * Standardized CORS, success, and error responses across backend operations.
  * Improved error responses by omitting sensitive details while preserving useful validation information.
  * Warehouse-transfer receipt creation now replaces existing receipt lines before adding generated lines.

* **Tests**
  * Added coverage for response formatting, CORS behavior, status codes, and error sanitization.

* **Chores**
  * Added commands for running tests and type checks without a lockfile.

### DIFF
--- a/packages/database/src/swagger-docs-schema.ts
+++ b/packages/database/src/swagger-docs-schema.ts
@@ -100190,7 +100190,7 @@ export default {
       properties: {
         id: {
           description:
-            "Note:\nThis is a Foreign Key to `supplierLocation.id`.<fk table='supplierLocation' column='id'/>",
+            "Note:\nThis is a Primary Key.<pk/>\nThis is a Foreign Key to `supplierLocation.id`.<fk table='supplierLocation' column='id'/>",
           format: "text",
           type: "string"
         },
@@ -100239,7 +100239,7 @@ export default {
         },
         supplierLocationId: {
           description:
-            "Note:\nThis is a Primary Key.<pk/>\nThis is a Foreign Key to `supplierLocation.id`.<fk table='supplierLocation' column='id'/>",
+            "Note:\nThis is a Foreign Key to `supplierLocation.id`.<fk table='supplierLocation' column='id'/>",
           format: "text",
           type: "string"
         },
--- a/packages/database/src/types.ts
+++ b/packages/database/src/types.ts
@@ -69811,14 +69811,14 @@ export type Database = {
           },
           {
             foreignKeyName: "address_countryCode_fkey"
-            columns: ["invoiceCountryCode"]
+            columns: ["customerCountryCode"]
             isOneToOne: false
             referencedRelation: "country"
             referencedColumns: ["alpha2"]
           },
           {
             foreignKeyName: "address_countryCode_fkey"
-            columns: ["customerCountryCode"]
+            columns: ["invoiceCountryCode"]
             isOneToOne: false
             referencedRelation: "country"
             referencedColumns: ["alpha2"]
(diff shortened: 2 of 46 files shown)


sidwebworks:
Addressed the CodeRabbit review:

**1. `lib/response.ts` `PostgresError` import (Critical) — false positive, no change.**
The web query analyzed npm `node-postgres` (`pg` 8.20.0), which exports `DatabaseError`. Carbon's `deno.json` maps `"pg"` → `https://deno.land/x/postgres@v0.17.0/mod.ts` (deno-postgres, a *different* package), which **does** export `PostgresError`. Verified by loading it directly:
```
$ deno run --config ./deno.json /tmp/pgcheck.ts
PostgresError is: function PostgresError
instanceof Error: true instanceof PostgresError: true
```
The module loads, and all 97 deno tests pass (including the `new PostgresError(...)` sanitizer case).

**2. embedding CORS preflight (Major) — fixed** in b8f6b4e7a. `corsPreflight` guard added. Note: current callers are the MCP tool (`shared_generateEmbedding`, server-side) and pg_net, neither of which sends a browser preflight, so this is defensive/consistency rather than an active break — but worth having.

**3. Generic safe `message` from the data-layer path (Major, 10 files) — intentional, no change.**
Omitting `message` (`{}`) for structurally-classified data-layer errors is a deliberate, load-bearing design, not a gap. The apps read the body via `getEdgeFunctionErrorMessage(err, fallback)` and every invoke site passes its own specific fallback; injecting a generic body message ("An error occurred") would *override* that better, caller-specific copy. The status is already preserved — `errorResponse(err, S)` keeps `S` (e.g. `isValidationError ? 400 : 500`) at every listed site. Authored `throw new Error(...)` messages still surface; only raw Postgres/Postgrest/node-pg/Zod detail is suppressed.

PR #1304, opened 2026-08-02
fix(inventory): link traceability storage units and de-duplicate transfer activity

sidwebworks:
## What does this PR do?

Three UI fixes to how stock transfers surface in inventory, all display-only — no schema change, no edge-function change, no migration. Traceability attributes rendered `To Shelf` with a raw storage-unit id; they now read **Storage Unit** (the domain term since the `shelf` rename in `20260417000100`) and resolve to the bin name, linking through to the quantities screen pre-filtered to that bin. The `Stock Transfer` attribute likewise rendered a raw id and now resolves to the readable id and links to the transfer, matching the existing Receipt/Shipment/Purchase Order renderers. Finally, the item activity feed renders one entry per `itemLedger` row, but a `Direct Transfer` writes two (out of the source, into the destination), so every transfer appeared twice — pairs are now collapsed into a single `from X to Y` entry.

**Before** — attribute panel showed `To Shelf` → `MlElhrU1l8QxQJInNZBVX` and `Stock Transfer` → `d9n73mkgk0h2lj3idd60`, both unlinked raw ids. The activity feed listed the same transfer twice: *"transferred 2 units"* and *"transferred 2 units to Dispatch counter"*.

**After** — `To Storage Unit` → **Dispatch counter** (links to the bin's quantities), `Stock Transfer` → **ST000001** (links to the transfer), and one activity entry per transfer.

> Screenshots captured against the local dev stack during verification; to be attached.

> Not ticked deliberately. Neither `erp` nor `@carbon/database` defines a `test` script, so there is no harness these would slot into. `collapseTransferPairs` is exported as a pure function and is the one genuinely unit-testable piece here if a harness is added later. Verification was done end-to-end against a running stack (below).

## How should this be tested?

Requires a running dev stack (`crbn up`) and a **posted** batch stock transfer — note that creating a transfer is not enough: the wizard only inserts `stockTransfer`/`stockTransferLine`, and `post-stock-transfer` (which writes the ledger rows and the `Transfer` activity) runs at pick time via the scan route. No environment variables; no migration.

1. **Attribute links** — open Traceability, search the batch number, select the lot, click its `Transfer` activity. The panel must show **Storage Unit** (not "Shelf") resolving to the bin *name*, and **Stock Transfer** resolving to the readable id. Both are links.
2. **Bin link target** — clicking the storage unit lands on Quantities with the filter chip *Storage Unit is any of &lt;bin&gt;* applied, scoped to the right location.
3. **Activity de-duplication** — open the item's Activity tab. One posted transfer must produce exactly **one** entry, reading `transferred N units from X to Y`. (The `from` clause is omitted when the source bin is null — a real case, since ledger rows can record stock against the location rather than a bin.)

Also worth exercising: a transfer whose source bin is null renders as *"No storage unit"* rather than a blank cell.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Inventory activity now combines direct stock transfers into a single, clearer entry showing both source and destination storage locations.
  - Transfer and storage-unit references in traceability details now display readable names with links to relevant inventory or detail views.
  - Legacy shelf attributes are presented using updated storage-unit labels and location context.

- **Bug Fixes**
  - Improved handling of transfer records to prevent duplicate outbound and inbound entries when both sides are available.
  - Added fallback labels when transfer or storage-unit details cannot be resolved.

### DIFF
--- a/apps/erp/app/modules/inventory/ui/Inventory/InventoryActivity.tsx
+++ b/apps/erp/app/modules/inventory/ui/Inventory/InventoryActivity.tsx
@@ -10,7 +10,60 @@ import Activity from "~/components/Activity";
 import { path } from "~/utils/path";
 import type { ItemLedger } from "../../types";
 
-const getActivityText = (ledgerRecord: ItemLedger) => {
+// A Direct Transfer writes two ledger rows in one transaction — the negative
+// out of the source bin and the positive into the destination — so the feed
+// renders the same move twice. Collapse each pair into the inbound row and hang
+// the source bin off it.
+//
+// The pair is identified by document + item + tracked entity + `createdAt`:
+// Postgres NOW() is transaction time, so both rows share it exactly. Keying on
+// the timestamp keeps repeat picks of the same line as separate entries, and
+// makes the merge idempotent, so it's safe to re-run over the accumulated list
+// as infinite scroll appends pages.
+export type CollapsedItemLedger = ItemLedger & {
+  transferFromStorageUnitName?: string | null;
+};
+
+export function collapseTransferPairs(
+  rows: ItemLedger[]
+): CollapsedItemLedger[] {
+  const pairKey = (row: ItemLedger) =>
+    [row.documentId, row.itemId, row.trackedEntityId ?? "", row.createdAt].join(
+      "|"
+    );
+
+  const outboundByKey = new Map<string, ItemLedger>();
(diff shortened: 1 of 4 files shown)


sidwebworks:
Superseded by #1306, which combines this with #1305 into a single review. Same commit (`ee87f114d`), cherry-picked onto the post-#1301 main.

PR #1305, opened 2026-08-02
fix(inventory): make the transfer wizard pull-based and add a lot picker to transfer scans

sidwebworks:
## What does this PR do?

Two follow-ups to the stock transfer wizard from #1301. First, the wizard now reads **pull-based**: a stock transfer starts from the bin that *needs* stock and pulls it from somewhere, so `Transfer To` is the master pane (step 1) and `Transfer From` is the detail (step 2) — the reverse of how #1301 shipped it. That inverts the delta preview, the shortage flag, the master ordering, the line mapping and the anchor of the over-allocation cap, not just the labels. Second, picking the tracked lot for a transfer line was scan-only; it now uses the shared `TrackedEntityPicker` (Scan **and** Select tabs) that picking lists and MES already use, which also fixes copy that said "serial number" for batch lines and the complete absence of expiry handling on that route. The drawer no longer renders the app-sidebar toggle.

**Before** — `Transfer From (1)` on the left, `Transfer To (2)` on the right, with a sidebar-collapse icon rendered inside the drawer header. The lot dialog was a single "Enter or scan serial number" field with Cancel/Pick and no way to browse.

**After** — `Transfer To (1)` → `Transfer From (2)`, no sidebar toggle. The lot dialog has **Scan | Select** tabs; Select lists each available lot with quantity and bin (`bPvS4u3eSWNHYmqjtB6YV · 2 available · Dispatch counter`) plus the FEFO/FIFO order control.

> Screenshots captured against the local dev stack during verification; to be attached.

> Not ticked deliberately — `erp` has no `test` script, so there is no harness these would slot into. Verified end-to-end against a running stack instead (below). Note this PR *deletes* ~120 lines of hand-rolled scan validation in favour of the shared picker, so the surface needing coverage shrinks.

## How should this be tested?

Requires a running dev stack (`crbn up`), a batch- or serial-tracked item with stock in at least two bins at one location, and a **Released** stock transfer with an unpicked tracked line. No environment variables; no migration.

1. **Pull direction** — open **Add Stock Transfer**. The left pane is `Transfer To`, ordered neediest-first. Select a destination; the right pane lists the other bins holding that item. Add one, then open **Review**: the line must read `<source bin> → qty → <destination bin>`, i.e. pulling *into* the row you selected.
2. **Over-allocation** — with one source feeding two destinations, the second destination's quantity must cap at what the source has left, not at its full availability.
3. **Sidebar toggle** — no panel/collapse icon in the drawer header.
4. **Lot picker** — open a tracked line's pick dialog. It must show **Scan | Select**; Select lists available lots with quantity and bin. For a batch item the copy must say "batch number", not "serial number". A lot already picked onto another line of the same transfer must not appear.

Verified against the local stack: Review showed `NO STORAGE UNIT → 3 → DISPATCH COUNTER` for a destination selected on the left, and the picker's Select tab listed three lots for a batch line.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Stock transfers now begin by selecting destination bins, then identify source bins to supply them.
  - Improved quantity guidance highlights destination shortages and respects available stock.
  - Batch and serial selection uses a streamlined picker that prevents duplicate selections.
  - Tables can optionally hide the sidebar trigger when displayed in drawers or modals.

- **Localization**
  - Updated inventory transfer terminology and prompts across supported languages.

### DIFF
--- a/apps/erp/app/components/Table/Table.tsx
+++ b/apps/erp/app/components/Table/Table.tsx
@@ -123,6 +123,9 @@ interface TableProps<T extends object> {
   withSavedView?: boolean;
   withSearch?: boolean;
   withSelectableRows?: boolean;
+  // Forwarded to TableHeader — hide the app-sidebar toggle when this Table is
+  // rendered inside a drawer or modal.
+  withSidebarTrigger?: boolean;
   withSimpleSorting?: boolean;
   sort?: ReactNode;
   getRowId?: (originalRow: T, index: number) => string;
@@ -272,6 +275,7 @@ const Table = <T extends object>({
   withSavedView = false,
   withSearch = true,
   withSelectableRows = false,
+  withSidebarTrigger = true,
   withSimpleSorting = true,
   sort,
   getRowId,
@@ -1068,6 +1072,7 @@ const Table = <T extends object>({
         withSavedView={withSavedView}
         withSearch={withSearch}
         withSelectableRows={withSelectableRows}
+        withSidebarTrigger={withSidebarTrigger}
         sort={sort}
       />
 
--- a/apps/erp/app/components/Table/components/TableHeader.tsx
+++ b/apps/erp/app/components/Table/components/TableHeader.tsx
@@ -88,6 +88,9 @@ type HeaderProps<T> = {
   setEditMode: (editMode: boolean) => void;
   table?: string;
   title?: string;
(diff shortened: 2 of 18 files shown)


sidwebworks:
Superseded by #1306, which combines this with #1304 into a single review. Both commits carried over unchanged.

PR #1332, opened 2026-08-05
feat(erp): bulk selection for BOM/BOP editors

sidgaikwad:
## Summary

- Adds **Select all / Deselect all / Delete (n)** controls to every BOM and Bill of Process editor — item make methods, job methods, and quote line methods. The controls are intentionally tiny and low-emphasis so they don't compete with the card actions, and they only render when the list is editable and non-empty.
- Fixes the mark-to-delete checkbox visibility: the checked state was a pale red check on the card background and was nearly invisible. It is now a red fill with a white checkmark in both light and dark mode.
- Bulk delete goes out as one request per card: the operation delete actions now accept multiple ids (single-id submissions still work), and three new bulk material delete routes cover method, job, and quote materials. Each surface keeps its existing guards — the revision lock per method row, the recorded-production-events check for job operations, and a single dependency/price recalculation after the batch.
- New strings are translated across all locales.

## Testing evidence

Verified end-to-end on a local dev environment:

- Select all / Deselect all toggle the whole list, with the delete button showing the selected count and each control disabling when it doesn't apply.
- Deleting a partial selection removes exactly those rows (confirmed in the database) and leaves the rest untouched.
- The controls hide on empty lists and read-only methods, and the same behavior was confirmed on the item, job, and quote editors.

BOM with all rows selected

## Summary by CodeRabbit

* **New Features**
  * Added bulk selection controls to bill-of-material and bill-of-process editors.
  * Users can select all, deselect all, and delete multiple materials or operations at once.
  * Added bulk deletion support across item methods, production jobs, and quotes.
  * Temporary entries are removed immediately, while saved records are processed securely.
* **Bug Fixes**
  * Added validation for missing selections, locked revisions, permissions, and production constraints.
  * Related pricing, dependencies, and ordering states are refreshed after deletion.
* **Style**
  * Improved deletion checkbox styling with clearer red status indicators.
* **Localization**
  * Added translated bulk-action labels across supported languages.

### DIFF
--- a/apps/erp/app/components/SortableList.tsx
+++ b/apps/erp/app/components/SortableList.tsx
@@ -121,7 +121,7 @@ function SortableListItem<T>({
                         id={`checkbox-${item.id}`}
                         aria-label={t`Mark to delete`}
                         onCheckedChange={() => onToggleItem(item.id)}
-                        className="border-foreground/20 bg-background/30 data-[state=checked]:bg-background data-[state=checked]:text-red-200 flex flex-shrink-0 "
+                        className="border-foreground/20 bg-background/30 data-[state=checked]:border-red-500 data-[state=checked]:bg-red-500 dark:data-[state=checked]:bg-red-500 data-[state=checked]:text-white flex flex-shrink-0"
                       />
                     )}
                     {/* List Order */}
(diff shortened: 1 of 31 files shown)


barbinbrad:
I'm closing this. It's too much screen real estate to optimize for a rare action.

PR #1334, opened 2026-08-05
Action bugs

aashu0148:
## What does this PR do?

- Adding a step to a job's Bill of Process no longer flickers. The whole form and the existing step list used to disappear and get replaced by a large spinner on every save, then pop back in. Now the "Save Step" button itself shows the spinner, the form empties out ready for the next step, and a "Step added" toast confirms it worked.
- Putting a huge rows-per-page value in the address bar (for example `?limit=100000`) no longer dumps every record onto a single page. It is now capped to the values in the "Results per page" dropdown (20 to 1000), and the page and the server read that same list so they cannot disagree.

## Summary by CodeRabbit

* **New Features**
  * Added a confirmation notification when a production step is successfully added.
  * New steps reset the form for quick entry of the next task.

* **Bug Fixes**
  * Prevented duplicate success notifications and preserved newly added items during saving.
  * Improved pagination handling for invalid, unsupported, or negative URL values.
  * Standardized available page-size options across tables.

* **Localization**
  * Added translations for step confirmations and bin-location matching messages across supported languages.


sidgaikwad:
lgtm, just fix the rabbit comments

sidgaikwad (approved):

PR #1339, opened 2026-08-05
feat(settings): derive business dates in company and location timezones

sidwebworks:
## What does this PR do?

**Problem.** Carbon answered "what day is it?" four different ways — UTC slicing (`toISOString().split("T")[0]`), the server's local zone (`today(getLocalTimeZone())`), the Deno runtime's zone, and Postgres `CURRENT_DATE` — all writing the same timezone-naive `DATE` columns. Production only worked because every runtime happens to default to UTC, and any non-UTC user's evening entries silently stamped the next calendar day: a PO created at 7pm in Chicago got tomorrow's `orderDate`, its journal could land in next month's accounting period, and `costLedger`/`itemLedger` rows for the same posting could disagree on the day (which mis-orders FIFO/LIFO cost layers).

**Design.** Business dates belong to a business calendar, not to whichever machine computed them. Two anchors, resolved **by domain**:

| Timezone | Owns | Why |
|---|---|---|
| `company.timezone` (new, required) | posting dates, accounting periods, aging, document-number date tokens, payroll/timecard weeks | one set of books = one calendar; subledger and GL must derive the same day |
| `location.timezone` (existing, required) | scheduling, shift rosters, MES day-grouping, expiry checks | "is this lot expired?" is answered at the plant, not at HQ |
| browser | display, pickers | the user's local time is correct for *showing* dates |

When a document has no location, location-scoped sites fall back to the company calendar. `TZ=UTC` is pinned across Dockerfiles/SST/compose so dev behaves like prod, and timestamps (`createdAt` etc.) stay plain UTC instants — only *calendar day* derivation changed.

**The pattern, per layer** — every "today" now takes an explicit IANA timezone:

- **Server TS** — `datetime.today(tz)` / `datetime.businessDay(instant, tz)` / `datetime.weekBounds(tz)` from `@carbon/utils`, with `tz` from `getCompanyTimeZone(client, companyId)` or `getLocationTimeZone(client, locationId, companyId)`. Routes use Redis-cached resolvers (`~/modules/shared/timezone.server`, invalidated when settings change); module services use the raw `@carbon/database` ones (module barrels leak into the client bundle, so no ioredis there — `build:erp` is the gate).
- **Edge functions** — same API from a Deno mirror (`functions/lib/datetime.ts`); resolvers accept a Supabase client or a Kysely `trx`.
- **SQL** — `company_today(companyId)` / `location_today(locationId, companyId)` replace `CURRENT_DATE` in every function that *writes* a business date (backflush, job completion, `get_next_sequence`, IC eliminations, shelf-life). Kept `SECURITY INVOKER` deliberately — a DEFINER version would be auto-exposed as a PostgREST RPC and let any user probe another tenant's local date.
- **Client forms** — a default that will be *persisted* as a business date (`orderDate`, `postingDate`, `dateIssued`…) seeds from `useCompanyToday()`, not the browser's day; intercompany posting dates follow the *source* company. Display-only code keeps using the browser zone.
- **Settings** — timezone is a required, searchable picker sourced from `pg_timezone_names` (so every value is valid for `AT TIME ZONE`), with DB `CHECK` constraints rejecting anything Postgres can't resolve.

**Keeping it fixed.** A `no-local-timezone` conformance check bans the whole bug class in server code — `getLocalTimeZone()`, UTC day-slicing, local date-parts of now (`new Date().getDay()`), and `setHours(0,0,0,0)` midnights. It scans services, jobs, edge functions, **and route loaders/actions** (route files are masked so their client half — components, hooks, `clientLoader`/`clientAction` — stays exempt). Dogfooding this check on the branch itself surfaced ~30 additional sites (shift rosters picking the server's weekday, payroll weeks on the server's calendar, new-company creation overwriting the user's chosen timezone with the server's) — all fixed here. A DST stress suite pins the hairy edges: spring-forward gaps and fall-back overlaps for wall-clock start times, zones that skip midnight itself (America/Santiago), 167/169-hour transition weeks, half-hour DST (Lord Howe), and why a fixed UTC cron can never express "02:30 local".

No linked issue — this came out of a timezone audit.

## How should this be tested?

**Migrations:** `pnpm db:migrate` (adds `company.timezone` backfilled from each company's first location, the `*_today()` helpers, tz-aware SQL functions, `get_timezone_names()`, validity CHECKs), then `pnpm run generate:types`.

**Automated:** `TZ=America/Chicago pnpm run test` is the real regression gate — the whole suite must stay green under a non-UTC process timezone. Also `pnpm --filter @carbon/utils test` (DST stress suite), `pnpm --filter @carbon/checks test` (conformance guard), and `pnpm run build:erp` (catches the ioredis client-bundle leak).

**Manual (differential-day test, playbook at `.ai/playbooks/company-timezone-differential.md`):** set the company timezone to one whose calendar day currently differs from UTC (`Pacific/Kiritimati`, UTC+14; confirm with `SELECT now()::date, company_today('<companyId>');`), then create a PO and post an inventory adjustment. `orderDate`, `itemLedger`/`costLedger`/journal `postingDate`, the document number's date tokens, and the lazily-created accounting period should all follow the *company's* day. `UPDATE company SET timezone='Fake/Zone'` must be rejected by the CHECK constraint.

## Summary by CodeRabbit

* **New Features**
  * Added company and location timezone settings with searchable timezone selection.
  * Added timezone-aware date handling across accounting, purchasing, sales, inventory, production, quality, and scheduling workflows.
  * Added timezone-aware sequence and serial-number previews.
* **Bug Fixes**
  * Improved business-date, posting-date, expiration, reporting, and weekly-period accuracy across time zones and daylight-saving transitions.
  * Added validation to prevent invalid timezone values.
* **Documentation**
  * Added guidance and testing procedures for timezone handling and date-related scenarios.


sidwebworks:
Addressed the CodeRabbit review in `021f33a6a` (pending push). Dispositions:

**Fixed**
- `timezone.server.ts` — location cache no longer stores inherited company timezones (an `inherit` sentinel + company-scoped key; company fallback flows through the invalidatable company cache), so `invalidateCompanyTimeZone` now takes effect immediately for inheriting locations.
- `packages/database/src/timezone.ts` + Deno mirror — resolvers throw on query failure instead of silently returning UTC; same for the MES expiry check's job lookup.
- `maskClientCode` — expression-bodied components no longer swallow subsequent server code. This was real in the wild: fixing it (plus tightening the closer so `) {` signature lines don't end regions) is covered by new regression tests.
- `no-local-timezone` — UTC-slicing pattern widened to argument-bearing constructors (`new Date(eventAt).toISOString().slice(0, 10)`); zero existing occurrences. `loadServerFiles` tolerates a missing root.
- `postJournalEntry` — the fallback posting date is persisted with the status flip.
- `issue` edge fn — accounting periods resolve from the same hoisted `today` as the ledger rows (`getAccountingPeriodForDate`), so a midnight rollover can't split journal from ledger.
- `generateEliminationEntries` — period resolved by date containment (`startDate <= company_today <= endDate`), Active-flag fallback kept for companies without a generated calendar.
- `company.timezone` backfill — filtered to `pg_timezone_names`-resolvable values.
- Intercompany sync keyed on `sourceCompanyId`; timecard week headers/day picker render company-calendar dates; schedule board client "today" uses the loader's location timezone; supplier-quote `expirationDate` is `parseDate`-validated; `getTimezones()` offset labels expire hourly; audit-archive key via `datetime.today("UTC")`; playbook thresholds + `companyId` scoping corrected; plan doc + AGENTS.md aligned.
- `extract-document` — reverted to local getters with the symmetry documented: only non-ISO text reaches that fallback, and `Date.parse` reads it in the process zone, so UTC getters (the earlier "fix") shifted the written date west of UTC. Good catch.

**Declined, with reasons**
- Rewriting `interpolateSequenceDate` internals / `isoWeekFromYmd` to `@internationalized/date`: `functions/lib/utils.ts` must stay free of `npm:` imports — it's reached by the node-side `@carbon/database` build, and pulling npm-specifier deps into that graph breaks it. The Intl-based derivation takes an explicit tz and is parity-tested against Postgres `IW`.
- `date-calculator.ts` rewrite: its native-Date usage is pure UTC arithmetic on date-only values (`Date.UTC` in, UTC getters out) — timezone-correct by construction; rewriting the scheduling engine's math is out of scope for this PR.
- Removing the 6 grandfathered `no-local-timezone` baseline entries: deliberate deferral, documented in the PR body and memory — each is a read-only/legacy site tracked for a follow-up.

sidwebworks:
Follow-up to the dispositions above: the first declined item is now **done** (`5e20d3b91`). Per review discussion, `@internationalized/date` gets the same treatment as `kysely` — a bare specifier mapped in the functions `deno.json` (npm: prefix dropped everywhere) and resolved from node_modules on the node side. That dissolved the constraint, so `isoWeekFromYmd` / `datetime.weekNumber` are now pure `CalendarDate` arithmetic (an oracle test pins parity with the old implementation for every day of 2019–2031) and `interpolateSequenceDate` reads its wall-clock parts from `now(timezone)` instead of `Intl.formatToParts`. Sequence tokens verified byte-identical between the Deno interpolator and SQL `get_next_sequence`. The other two declines (date-calculator rewrite, baseline entries) stand.

PR #1345, opened 2026-08-07
fix: recover from orphaned auth user on re-creation after partial delete

aashu0148:
What does this PR do?

- Fixes a bug where deleting a user and then trying to recreate them with the same email showed a "user already exists" error. When a user was deleted, two cleanup operations ran at the same time: if one succeeded and the other failed, the account ended up in a broken half-deleted state that permanently blocked recreation.
- The deletion is now done in two sequential steps so a failure leaves the account fully intact and retryable, rather than partially destroyed.
- If a user is in the broken state already, recreation now automatically detects and recovers from it instead of failing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account creation for customer, employee, and supplier accounts by recovering existing authentication accounts when possible.
  * Added clearer handling when authentication account setup fails.
  * Improved account deletion reliability by ensuring authentication records are removed before application records.

### DIFF
--- a/apps/erp/app/modules/users/users.server.ts
+++ b/apps/erp/app/modules/users/users.server.ts
@@ -270,17 +270,12 @@ export async function createCustomerAccount(
     userId = user.data.id;
   } else {
     isNewUser = true;
-    const createSupabaseUser = await serviceRole.auth.admin.createUser({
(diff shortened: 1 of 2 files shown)


naveenkash (approved):
LGTM

PR #1369, opened 2026-08-11
fix(erp): trim whitespace on validator name and code fields

BettrCallRaul:
## What

`.trim()` before `.min(1)` on 70 `name` and 6 `code` fields across 15 ERP `*.models.ts`. Nothing else changes. No `id` fields touched — those are machine-supplied (hidden edit field, or Combobox selections of existing rows), so whitespace can't reach them.

## The issue

Reported on Inventory › Storage Types: trailing spaces are kept as part of the name. The validator was

```ts
name: z.string().min(1, { message: "Name is required" })
```

`min(1)` counts whitespace, so `"Pallet "` passes and is stored verbatim. Nothing downstream trims it, and the trailing space is invisible in every rendered view. This was the codebase default, not a one-off — the same pattern was on 76 fields, with one pre-existing counter-example (`closeTaskSkipValidator.skippedReason`).

## The risk

**1. It bypasses uniqueness constraints.** ~55 tables carry `UNIQUE (name, "companyId")`. Postgres compares the raw string, so a leading or trailing space defeats it.

Worked example, departments (`UNIQUE (name, "companyId")`):

| Step | Result |
|---|---|
| `Quality` already exists | 1 department |
| User types `"  Quality  "` | passes `min(1)` |
| DB compares `"  Quality  "` ≠ `"Quality"` | constraint does **not** fire |
| Row inserts | **2 departments, both displaying "Quality"** |

From there the two are indistinguishable in the UI: the Department picker lists "Quality" twice with nothing to tell them apart, employees get assigned to either, and anything grouped by department splits across both. Recovering means merging records by hand. The constraint exists precisely to prevent this.

**2. Whitespace-only names are accepted.** `"   "` satisfies `min(1)`, creating a record whose name renders as blank.

**3. `code` fields feed generated identifiers.** `materialForm` / `materialSubstance` / `materialType` / `unitOfMeasure` codes are composed into material IDs, so whitespace corrupts a derived key rather than just a label. Trimming before `.max(10)` also makes that length check measure the real value.

## Verification

Local instance, checking stored bytes in Postgres (a badge renders `Pallet` and `Pallet ` identically, so the UI can't confirm this):

| Input | Before | After |
|---|---|---|
| `"   PALLET   "` | stored with spaces | stored `[PALLET]`, length 6 |
| `"     "` | row created, blank name | rejected, "Name is required" |
| `"  Quality  "`, `Quality` exists | 2nd department created | rejected by the unique constraint |

New `inventory.models.test.ts` (4 tests). Typecheck clean. Suite: 513 passed, 5 pre-existing failures confirmed identical on `main`.

## Excluded

`configurationRuleValidator.code` (JavaScript source) and `oAuthCallbackSchema.code` (OAuth code).

## One unrelated change

`items.models.ts` imports shared enums from `../shared/shared.models` rather than the `../shared` barrel, matching `quality.models.ts`. The barrel pulls in `@carbon/glossary`, whose Lingui macros aren't transformed under vitest, so tests importing it failed at import.

## Not addressed

- MCP/API writes go through services, not validators — still untrimmed.
- `storageType` has no unique constraint at all, so duplicates remain possible there regardless.
- `zfd.text()` doesn't trim; it only maps an empty string to `undefined`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation across ERP forms by automatically removing leading and trailing whitespace from names, codes, and document names.
  - Entries containing only whitespace are now rejected consistently while existing validation rules remain unchanged.

- **Tests**
  - Added coverage confirming storage types and units trim accepted names and reject whitespace-only values.

### DIFF
--- a/apps/erp/app/modules/accounting/accounting.models.ts
+++ b/apps/erp/app/modules/accounting/accounting.models.ts
@@ -277,7 +277,7 @@ export const reportViewValidator = z.object({
 export const groupAccountValidator = z
   .object({
     id: zfd.text(z.string().optional()),
-    name: z.string().min(1, { message: "Name is required" }),
+    name: z.string().trim().min(1, { message: "Name is required" }),
     parentId: zfd.text(z.string().optional()),
     accountType: z
       .enum(accountTypes, {
@@ -331,7 +331,7 @@ export const accountValidator = z
   .object({
     id: zfd.text(z.string().optional()),
     number: z.string().min(1, { message: "Number is required" }).nullish(),
-    name: z.string().min(1, { message: "Name is required" }),
+    name: z.string().trim().min(1, { message: "Name is required" }),
     parentId: zfd.text(z.string().optional()),
     isGroup: zfd.checkbox(),
     accountType: z
@@ -415,7 +415,7 @@ export const journalLineValidator = z.object({
 
 export const currencyValidator = z.object({
   id: zfd.text(z.string().optional()),
-  code: z.string().min(1, { message: "Code is required" }),
+  code: z.string().trim().min(1, { message: "Code is required" }),
   decimalPlaces: zfd.numeric(z.number().min(0).max(4)),
   exchangeRate: zfd.numeric(z.number().min(0, { message: "Rate is required" })),
   historicalExchangeRate: zfd.numeric(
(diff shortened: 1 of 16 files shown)


sidwebworks (approved):
LGTM

PR #1370, opened 2026-08-11
Fix/bug fixes 11aug

aashu0148:
## What does this PR do?

- Editing a supplier price through the API no longer fails as a duplicate. Every API call was being treated as "create a new record" even when it was an edit, so the edit collided with the record it was trying to change.
- The page loading bar at the top now has a moving shine, so a slow page change reads as progress instead of a frozen bar. Safari does not pre-load pages the way Chrome does, so switching tabs there can sit for several seconds with no sign anything is happening.
- The production planning "create jobs" bulk action now lists the jobs it created and says why it skipped an item (an open job already covers it, or there is nothing to make), instead of lumping skips in with failures.

## video demo

https://github.com/user-attachments/assets/48ced95c-ba61-4dc3-9b2d-ca9dca405f77

## Summary by CodeRabbit

* **New Features**
  * Production planning results now show created and updated job counts, skipped items, job links, and a “View all” option.
  * Items without orders are clearly identified and excluded from processing.
* **Bug Fixes**
  * Supplier-part edits no longer overwrite creator information or get treated as new records.
  * Unnecessary job-status tooltip text has been removed.
* **Style**
  * Updated the loading progress indicator with a blue animated design and reduced-motion support.
* **Localization**
  * Added and updated production-result messages across supported languages.


naveenkash (approved):
lgtm

sidwebworks (commented):
Are there any other places with [this same issue](https://github.com/crbnos/carbon/pull/1370/changes#diff-2320c7ca4c10589312ba9cbedc9db0017415e3d040bd96c5699ae388bb39b257R3795-R3802)? I think so there are - shouldn't we handle this for other paths too or a shared util.

PR #1378, opened 2026-08-12
fix(itar): exempt Carbon staff from the customer entity gate

chasefostermfg:
## What does this PR do?

Follow-up to #1377. Provisioning a customer company in GovCloud put **us** in front of the customer's entity acceptance screen, with no way past it.

`seed-company` gives the user who creates a company the **Admin** employee type in it, which includes `users_update` — and the entity screen keys on exactly that permission. So "add a company" made the Carbon staffer an admin of the customer's tenant, and the gate then asked them to accept the Carbon GovCloud Rider on the customer's behalf: a document only the customer can sign.

The fix adds `requiresItarEntityCertification(email)` to `packages/utils/src/const.ts`, next to `isInternalEmail` (the same helper that already gates company creation and backs `isCarbonOwnedCompany`), and uses it to drop the **entity** half of the gate for internal accounts:

- **Layouts (ERP + MES)** compute `entityRequired` **server-side** from the account's email. Carbon staff skip both the entity screen and the pending block and land straight on their own U.S.-Person attestation.
- **Both acknowledge actions** refuse `intent=itar-entity` from a staff account, so the hidden screen is not the only thing standing between a staff session and binding a customer. They also refuse callers with no signer identity — the API-key auth path reports an empty email, and a Rider acceptance is a named person signing.
- **The compliance report** gains the company-level half it was missing: an entity-status banner reading Accepted / Expired / Pending, with the signer, Rider version and expiry. Pending is the normal state for a freshly provisioned tenant, and stays that way until the customer's own first admin accepts.

Unchanged on the customer side: their admin (`users_update`) gets the entity screen, everyone else waits on the pending block. The **user** attestation still applies to every account per company, staff included — so a customer's ITAR certifications report shows a current attestation for the Carbon staff in their tenant, exactly as it does for their own people.

Net effect: add a company and invite the customer's admin exactly as before, and only the customer's first admin ever binds the customer.

### Two calls worth a reviewer's eye

- **API-key callers are now refused on the entity intent too.** That path authenticates as the key and carries no signer identity. It is a pre-existing surface rather than part of this bug, but closing the staff route while leaving it open seemed wrong. Two lines; cannot affect the browser flow.
- **Staff skip the entity screen in every company, including a Carbon-owned tenant.** This reads the Rider as Carbon↔customer — Carbon does not sign one with itself — so a Carbon tenant on GovCloud would show "entity: pending" indefinitely. If such a tenant genuinely needs its own entity cert, that needs a signal beyond the email domain.

No browser walkthrough — the gate only renders under `CONTROLLED_ENVIRONMENT=true` and there was no running controlled stack available. Manual steps are in "How should this be tested?" below.

- [x] I confirm automated tests are in place that prove my fix is effective or that my feature works. (Unit tests on the exemption predicate, including the near-miss addresses that must not receive it; the gate wiring itself is covered by scoped typecheck + manual steps below.)

## How should this be tested?

Env: `CONTROLLED_ENVIRONMENT=true`. No migration and no schema change — `getItarCertificationStatus` is untouched; `entityRequired` is layered on in the loaders.

**The reported bug (ERP):** as an `@carbon.ms` account, create a company in a controlled environment and enter it. Expected: no entity screen and no pending block — straight to the U.S.-Person attestation, then into the app. Before this change, the entity Rider screen appeared and could not be passed.

**Customer side is unchanged:** invite an admin (`users_update`) into that fresh company. They get the entity screen; a non-admin gets the pending block. Once the admin accepts, everyone falls through to their own attestation.

**MES:** same two accounts. Staff go straight to the attestation; the customer's non-admin sees the pending block until an admin accepts in the ERP.

**Server-side enforcement:** POST `intent=itar-entity` to `/x/acknowledge` (ERP and MES) from a staff session holding `users_update`. Expected: refused with "The Rider must be accepted by a representative of the company it binds", and no `itarCertification` row of type `entity` written. Same request with a `carbon-key` header is refused for the same reason.

**Report:** `/x/settings/itar-certifications` on the fresh company shows "Entity certification: Pending" above the table, and the staff account appears in the table with a current attestation. After the customer's admin accepts, the banner flips to Accepted with their name, title, Rider version and expiry. Backdate the entity row's `expiresAt` to see Expired.

**Automated:**

```bash
pnpm exec turbo run test --filter=@carbon/utils      # 138 passed (5 new)
pnpm test                                            # 23/23 packages green
pnpm exec turbo run typecheck --filter=erp --filter=mes --filter=@carbon/utils
```

Also run: 6 new UI strings extracted and translated across all 12 non-English catalogs (0 missing), and `pnpm generate:mcp` re-run for the new `users_getItarEntityCertification` service function.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01VhnanCNS8XMTvJKGb9pJhi)_

## Summary by CodeRabbit

- **New Features**
  - Added company-level ITAR certification status displays for accepted, expired, and pending states.
  - Settings now show signer details, acceptance and expiration dates, and Rider version information when available.
  - Added certification visibility across ERP and MES experiences.
- **Access Controls**
  - ITAR access now enforces entity certification where required.
  - Only eligible company representatives can submit entity certifications.
  - Internal staff exemptions remain supported.
- **Localization**
  - Added translated certification, expiration, access, and Rider messaging across supported languages.

### DIFF
--- a/apps/erp/app/modules/settings/ui/ItarCertifications/ItarEntityCertificationStatus.tsx
+++ b/apps/erp/app/modules/settings/ui/ItarCertifications/ItarEntityCertificationStatus.tsx
@@ -1,0 +1,90 @@
+import { Alert, AlertDescription, AlertTitle } from "@carbon/react";
+import { formatDate } from "@carbon/utils";
+import { Trans, useLingui } from "@lingui/react/macro";
+import { LuBuilding2, LuCircleAlert, LuCircleCheck } from "react-icons/lu";
+
+export type ItarEntityCertificationRecord = {
+  userId: string | null;
+  fullLegalName: string | null;
+  title: string | null;
+  complianceContact: string | null;
+  docVersion: string | null;
+  certifiedAt: string | null;
+  expiresAt: string | null;
+};
+
+type ItarEntityCertificationStatusProps = {
+  certification: ItarEntityCertificationRecord | null;
+};
+
+/**
+ * Company-level half of the compliance report: whether a representative of this
+ * company has accepted the Carbon GovCloud Rider. The table below it is
+ * per-person (U.S.-Person attestations); this is a single fact about the entity,
+ * so it reads as a status banner rather than a column repeated on every row.
+ *
+ * Pending is the normal state for a freshly provisioned tenant — Carbon staff
+ * set the company up but cannot sign on its behalf, so it stays pending until
+ * the company's own first admin accepts.
+ */
+const ItarEntityCertificationStatus = ({
+  certification
+}: ItarEntityCertificationStatusProps) => {
+  const { t } = useLingui();
+
+  const expiresAt = certification?.expiresAt ?? null;
+  const isExpired = expiresAt !== null && Date.parse(expiresAt) <= Date.now();
+  const isCertified = certification !== null && !isExpired;
+
+  return (
+    <div className="w-full px-4 pt-4">
+      <Alert variant={isCertified ? "success" : "warning"}>
+        {isCertified ? <LuCircleCheck /> : <LuCircleAlert />}
+        <AlertTitle>
+          {isCertified ? (
+            <Trans>Entity certification: Accepted</Trans>
+          ) : isExpired ? (
+            <Trans>Entity certification: Expired</Trans>
+          ) : (
+            <Trans>Entity certification: Pending</Trans>
+          )}
+        </AlertTitle>
+        <AlertDescription>
+          {certification ? (
+            <span className="flex flex-wrap gap-x-4 gap-y-1">
+              <span className="inline-flex items-center gap-1.5">
+                <LuBuilding2 className="size-3.5 shrink-0" />
+                {certification.fullLegalName}
+                {certification.title ? `, ${certification.title}` : null}
+              </span>
+              {certification.certifiedAt ? (
+                <span>
+                  {t`Accepted`} {formatDate(certification.certifiedAt)}
+                </span>
+              ) : null}
+              {certification.docVersion ? (
+                <span>
+                  {t`Rider`} v{certification.docVersion}
+                </span>
+              ) : null}
+              {expiresAt ? (
+                <span>
+                  {isExpired ? t`Expired` : t`Expires`} {formatDate(expiresAt)}
+                </span>
+              ) : null}
+            </span>
+          ) : (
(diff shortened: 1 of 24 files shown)


chasefostermfg:
All 10 review comments are addressed in d0ade083. Two were fixed at their root rather than literally, so noting the reasoning here rather than on each thread.

**Discarded query error** (`itar-certifications.tsx`) — correct, and the worst possible failure mode for this page. A missing entity row and a failed entity query both arrive as "no data" but mean opposite things: a company that hasn't signed the Rider, versus a company whose status we don't know. The banner reported the second as "Pending", which puts a false compliance finding in front of an auditor. Both reads now fail the page with a flash instead. I fixed `report.error` the same way while I was in there — it had the identical discard (pre-existing from #1377), so a failed employee read rendered an empty table as though nobody was certified.

**Expiry comparison** — applied as suggested; `parseAbsolute(expiresAt, "UTC").compare(now("UTC")) <= 0`.

**`Accepted` gender in it/pt** — the finding was right, but the cause was upstream of the translation: I had assembled the description from fragments (`t\`Accepted\`` + a formatted date), which hands translators a bare adjective with no subject to agree with. Rather than gender the fragment, the strings are now whole phrases — `Accepted {0}`, `Expires {0}`, `Rider v{0}` — so agreement is expressible in every locale. Those two msgids no longer exist. The expired case turned out to reuse the catalog's existing `Expired {0}`, so it's one fewer new string than before.

**"An administrator here" / access wording in de, es, it, pl** — four locales independently flagged the same sentence, which pointed at the English source rather than the translations. Changed to "An administrator **at this company** must accept it before anyone can **access Carbon**", then retranslated all 12 locales from the clearer text, including the specific wording each comment suggested (`Administrator dieses Unternehmens` / `Zugriff auf Carbon erhält`, `de esta empresa` / `acceder`, `dell'azienda` / `accedere`, and Polish's informal address form with `uzyskać dostęp`).

**Glossary terms** — applied as suggested: it `In sospeso`, ru `Ожидание`.

Verification: `pnpm test` 23/23 packages, typecheck clean for erp/mes/utils, biome clean, 0 missing translations across all 13 catalogs.

---
_Generated by [Claude Code](https://claude.ai/code)_

barbinbrad (approved):

PR #1381, opened 2026-08-12
[superseded by #1382] feat(items): item rules for sales documents, with enforcement gates

naveenkash:
Configurable **if-condition-then-error/warn rules** that fire when an item is added to a quote or sales order line — e.g. "if the ship-to country is X, block this part." Built on the existing storage-rules engine rather than duplicating it.

Tracking spec: `.ai/specs/2026-08-11-item-rules.md` · audit: `.ai/research/2026-08-12-item-rules-enforcement-surface.md` (docs PR #1368)

## Feature

- Three additive migrations: `itemRule`, `itemRuleAssignment`, `itemRuleAcknowledgment` (append-only override/block evidence), plus a `companySettings` notification group.
- A **separate** `itemRuleSurface` enum rather than extending the storage `transactionSurface`. The two families fire on categorically different things, and separate enums make "a storage rule subscribed to a sales surface" unrepresentable instead of merely discouraged.
- Engine extensions in `@carbon/utils`: a `customer` context root (type, status, ship-to country as alpha-2), a separate `ITEM_RULE_FIELD_REGISTRY`, and a shared zod AST mirror in `rules-schema.ts` so neither ERP module has to import the other.
- Evaluator in `@carbon/ee` — service-role, plan-gated on `ITEM_RULES`, with explicit `companyId` predicates on every query since RLS is bypassed.
- Admin UI under Items, per-item assignment, notifications, and the shared violation modal.

## Refactor (same PR, because the feature forces it)

Item rules **share** the storage-rules engine, evaluator structure, and violation modal rather than duplicating them. A sales feature importing from `storage-rules`-named modules would read as a dependency on the storage *feature*, so the shared code moved to neutral `rules` naming: `utils/storage-rules.ts` → `rules.ts`, `ee/{storage-rules,item-rules}` → `ee/rules/{storage,item}` with the modal and hook hoisted to `rules/`. Exports are `./rules` + `./rules.server`. **Tables stay separate — only the code layer is unified.**

Module placement also fixed: a rule feature isn't its own domain, so neither gets a standalone `~/modules/*`. Storage rules folded into `~/modules/inventory`, item rules into `~/modules/items`.

## Enforcement

Per-line checks alone can't hold this. 14 entry points can put an `itemId` on a sales line and only 4 were enforced — conversions, quote duplication, CAD drag-drop, the Paperless Parts webhook and MCP all write lines without them.

So enforcement is now **per-gate**, not only per-write: terminal gates that re-read the whole document at **sales order confirm, quote finalize, quote→order convert, RFQ→quote convert, and shipment post**, copying the pattern already running at receipt/shipment posting. A gate covers writers nobody instrumented *and* catches staleness (a rule authored after the line, a ship-to that changed). The customer-facing digital-quote accept is **hard-block-on-error only** — there's no employee there to acknowledge, and internal compliance text must not reach the customer. An MCP backstop covers the two sales line writers; `sales_insertSalesOrderLines` is blocklisted (it has no in-app caller).

## Fixes

- **Drop shipments evaluated the wrong address** — the order header's ship-to rather than `salesOrderShipment.customerLocationId`, so a country rule could clear an order shipping somewhere else entirely.
- **Quote lines evaluated a hardcoded quantity of 1**, so quantity rules never fired on quotes. Now uses the largest quantity break.
- **The evaluator selected `item.customFields`, a column that doesn't exist.** PostgREST failed the whole query, the error was never checked, and zero items came back — which silently skipped *every broadcast rule*. Removed the column; a failed item load now throws rather than quietly disabling enforcement.

## Verification

Typecheck 7/7 touched packages · biome clean · `@carbon/utils` 76 tests, `@carbon/ee` 50, `mes` 47, `erp` 246 passing · build green · 132 translations filled across 12 locales (`linguito check` exit 0).

Two ERP test files fail (`i18n-react-macros`, `localized-submodule-ui`) — **proven pre-existing**: forbidden-pattern counts are byte-identical to `HEAD` for the two files this PR touches, and the remaining 15 flagged files are untouched here.

## Known gaps (deliberate, documented in the audit)

Direct PostgREST writes with an API key can't be caught at the app layer; standalone sales invoices reach revenue with no upstream document; service-only orders skip shipment so they miss that checkpoint; a nightly re-scan and a Postgres-trigger backstop were both considered and declined. Unifying the two rule tables is a separate follow-up.

**Not browser-tested end to end** beyond the RFQ→quote and quote-line paths exercised while fixing the bugs above.

## Summary by CodeRabbit

* **New Features**
  * Added item-rule management, filtering, assignment, severity, and surface configuration.
  * Added item-rule enforcement for quote and sales-order lines, conversions, confirmations, digital acceptance, and shipment posting.
  * Added acknowledgment workflows, violation details, configurable notifications, and document links.
  * Added customer-aware rule conditions and localized interface text.
  * Added MCP safeguards for unrestricted sales-line insertion.
* **Bug Fixes**
  * Unified rule-violation handling across inventory, sales, shipment, and transfer workflows.
* **Documentation**
  * Added guidance and API documentation for item rules and rule configuration.


naveenkash:
Superseded by #1382 — same commits, same diff. Closed automatically when the branch was renamed.

PR #1382, opened 2026-08-12
feat(sales): sales rules on a unified enforcementRule table, with enforcement gates

naveenkash:
Configurable **if-condition-then-error/warn rules** that fire when an item is added to a quote, sales order, or sales invoice line — e.g. "if the ship-to country is X, block this part." Built on the existing storage-rules engine rather than duplicating it.

Tracking spec: `.ai/specs/2026-08-11-item-rules.md` · audit: `.ai/research/2026-08-12-item-rules-enforcement-surface.md` (docs PR #1368)

## Feature

- **One rule table for both families.** `enforcementRule` carries storage rules and sales rules, discriminated by a `family` column (`'storage' | 'sales'`), with `enforcementRuleItemAssignment` (both families), `enforcementRuleWorkCenterAssignment` (storage only), and `enforcementRuleAcknowledgment` (append-only override/block evidence). Two migrations: create the unified schema, then move the shipped storage rules onto it and drop the old tables.
- **Cross-family subscription is blocked by CHECK, not by column type.** One `enforcementRuleSurface` enum spans both families, so `enforcementRule_storage_surfaces` / `enforcementRule_sales_surfaces` keep a storage rule from subscribing to `quoteLine` (a rule that would silently never fire), and `enforcementRule_sales_shape` pins sales rows to `targetType='item'`. That swap — structural isolation becomes constraint + query isolation — is the deliberate cost of the single table, and `family-isolation.test.ts` holds the query half.
- Engine extensions in `@carbon/utils`: a `customer` context root (type, status, ship-to country as alpha-2), a `SALES_RULE_FIELD_REGISTRY`, and a shared zod AST mirror in `rules-schema.ts` so neither ERP module has to import the other.
- Evaluator in `@carbon/ee` — service-role, plan-gated on `SALES_RULES`, with explicit `companyId` predicates on every query since RLS is bypassed.
- Admin UI under **Sales**, per-item assignment, notifications, and the shared violation modal.

## Refactor (same PR, because the feature forces it)

Sales rules **share** the storage-rules engine, evaluator structure, and violation modal rather than duplicating them. A sales feature importing from `storage-rules`-named modules would read as a dependency on the storage *feature*, so the shared code moved to neutral `rules` naming: `utils/storage-rules.ts` → `rules.ts`, `ee/storage-rules` → `ee/rules/{storage,sales}` with the modal and hook hoisted to `rules/`. Exports are `./rules` + `./rules.server`.

Module placement also fixed: a rule feature isn't its own domain, so neither gets a standalone `~/modules/*`. Storage rules folded into `~/modules/inventory`, sales rules into `~/modules/sales`. The admin CRUD is written **once**, parametrized by family, in `~/modules/shared` — callers pass `"storage"` or `"sales"`.

## Enforcement

Per-line checks alone can't hold this. 14 entry points can put an `itemId` on a sales line and only 4 were enforced — conversions, quote duplication, CAD drag-drop, the Paperless Parts webhook and MCP all write lines without them.

So enforcement is **per-gate**, not only per-write: terminal gates that re-read the whole document at **sales order confirm, quote finalize, quote→order convert, RFQ→quote convert, and shipment post**, copying the pattern already running at receipt/shipment posting. A gate covers writers nobody instrumented *and* catches staleness (a rule authored after the line, a ship-to that changed). The customer-facing digital-quote accept is **hard-block-on-error only** — there's no employee there to acknowledge, and internal compliance text must not reach the customer. An MCP backstop covers the three sales line writers **and** the terminal transitions (`sales_finalizeQuote`, `sales_convertQuoteToOrder`, `sales_convertSalesRfqToQuote`) — error-severity blocks, warns pass since no human is present to acknowledge; `sales_insertSalesOrderLines` is blocklisted (it has no in-app caller).

### Sales invoices (implemented — closes the "skip the order" bypass)

Every gate above sits on the quote→order→shipment path, but Carbon lets a user raise a sales invoice with **no upstream document at all** — so a restricted item invoiced directly would reach revenue having passed nothing. This PR closes that: a `salesInvoiceLine` surface joining the other two (enum value + CHECK widening in two migrations, since `ALTER TYPE ADD VALUE` can't be referenced in its own transaction), line checks with acknowledgment evidence + notifications on the two invoice-line write routes, and a terminal gate in the `$invoiceId.post` route action **before** the Stripe preflight and the optimistic `status='Pending'` write, so a blocked post touches nothing (void is never gated — it's corrective).

The design problem is that an invoice has **no customer ship-to** (`invoiceCustomerLocationId` is the bill-to; the `locationId` columns are Carbon's own warehouses). It resolves cleanly: an order-derived line (`salesOrderId` set) resolves the ship-to through its source order via `resolveSalesOrderShipTo` — the document gate groups lines **per source order** — drop-ship handling included; a **standalone** line passes `customerLocationId: null` and the engine's required-field semantics fire — any rule referencing the ship-to country raises "Customer country is required" at that rule's severity. The bypass path carries strictly less information than the path it skips, so it **fails closed**, and rules that never mention the ship-to evaluate at full strength. The bill-to is never substituted — it's frequently a different country, the same failure shape as the drop-ship defect fixed below — and `invoice-shipto.test.ts` pins exactly that: the standalone-line test seeds a safe-country bill-to on the invoice so a future "fix" that substitutes it fails loudly. The MCP backstop gains `invoicing_upsertSalesInvoiceLine` → `salesInvoiceLine` (errors block, warns pass; an update of an order-derived line resolves its real destination); the check lives in the executor rather than inside `upsertSalesInvoiceLine` for the same reason as the other two — service files are client-bundled via module barrels and must stay free of server-only imports.

## Fixes

- **Drop shipments evaluated the wrong address** — the order header's ship-to rather than `salesOrderShipment.customerLocationId`, so a country rule could clear an order shipping somewhere else entirely. When a drop shipment has no location the resolver now returns null rather than falling back to the header, so the engine's required-field semantics surface it instead of silently checking a different country.
- **Quote lines evaluated a hardcoded quantity of 1**, so quantity rules never fired on quotes. Now uses the largest quantity break.
- **The evaluator selected `item.customFields`, a column that doesn't exist.** PostgREST failed the whole query, the error was never checked, and zero items came back — which silently skipped *every broadcast rule*. Removed the column; a failed item load, rule load, or ship-to resolution now throws rather than quietly disabling enforcement.

## Review focus

The **RLS policies** are the part worth reading hardest. Writes on `enforcementRule` are gated per family (storage → `inventory_*`, sales → `sales_*`, with UPDATE guarded by both `USING` and `WITH CHECK` so a row can't be edited across families), and the shared item-pin table resolves the pinned rule's family through an `EXISTS` so each family keeps exactly the permission its source table required — the merge changes no caller's authorization.

Note the correlation in that `EXISTS` is qualified by table name on purpose: an unqualified `"companyId"` binds to the inner rule table and turns the tenant check into a tautology. `"ruleId"` needs no qualification because `enforcementRule` has no such column, which is what makes the mistake easy to miss.

## Self-review round (post-review fixes, same PR)

A full self-review pass over the branch produced a follow-up round of fixes:

- **MCP terminal-gate bypass closed.** The three MCP-exposed terminal transitions above ran no evaluation at all — an agent with an OAuth token could finalize or convert a quote carrying error-severity violations. They now run `evaluateSalesRulesForSalesDocument` in the executor.
- **Override evidence at the gates.** Only the per-line routes wrote `enforcementRuleAcknowledgment` rows; the strongest overrides (confirming, posting, shipping past an acknowledged warn) left no trail. The evidence write + notification is now one helper (`recordSalesRuleOutcome` in `sales.server.ts`) called from all six line routes, the five recordable gates, and the MCP backstops' blocks. RFQ convert deliberately records nothing (the table's `documentType` CHECK has no `salesRfq`; the minted quote re-records downstream).
- **Tenant scoping.** The evaluator's `customerLocation` ship-to read gained an explicit `companyId` predicate (the invoice-post gate evaluates with the service-role client, so RLS was not backstopping it), and the shared `getEnforcementRule`/`upsertEnforcementRule`/`deleteEnforcementRule` now require `companyId` — defense-in-depth for the MCP-exposed CRUD.
- **UI correctness.** The four rule-save routes no longer discard the flash cookie on failure; `useRuleViolations` no longer fires `onSuccess` for `{ success: false }` payloads (confirm/finalize modals stayed closed over real failures); dead header fetchers removed; storage-rule builders no longer fetch customer option lists they never show.
- **Hygiene.** ~17 storage-family comments/identifiers mis-renamed to "sales rules" corrected; three dead validators removed; storage-direction family-isolation tests added; the stale generated swagger schema (still carrying `storageRule*` definitions) committed.
- **Docs.** New reader-facing reference page at `docs/reference/sales-rules` (+ agent KB regen, glossary entry, cross-links), AGENTS.md freshness fixes, and the implemented spec moved to `.ai/specs/implemented/`.

## Verification

Typecheck 7/7 touched packages · biome clean · `@carbon/utils` 183, `@carbon/ee` 586 (incl. the invoice ship-to suite and both directions of family isolation), `@carbon/jobs` 490 tests passing · erp production build green · docs build green · `linguito check` exit 0. (Five erp test failures are pre-existing — reproduced identically with this branch's changes stashed.)

The migrations were applied from scratch against a clean database (948 migrations, no errors), the CHECK constraints were exercised against the illegal states they exist to reject, and the data migration was run against seeded rows to confirm the nullable filter trio normalizes, the surface array casts, work-center targets survive, and a re-run is a no-op.

## Known gaps (deliberate, documented in the audit)

Direct PostgREST writes with an API key can't be caught at the app layer; a service-role caller invoking the `post-sales-invoice` edge function directly bypasses the invoice-post gate (same residual hole as the convert gates — the MCP error-severity backstop is what covers the line writes on that path); a nightly re-scan and a Postgres-trigger backstop were both considered and declined. `acknowledged` is client-supplied, so a crafted first submit can skip a `warn` (errors block unconditionally) — though the acknowledgment evidence row is written either way.

**The RLS policies have not been exercised by a real authenticated request** — typecheck, tests, and a schema diff all pass on a policy whose tenant correlation does nothing, so this needs a logged-in user before merge. Not browser-tested end to end beyond the RFQ→quote and quote-line paths exercised while fixing the bugs above.

## Summary by CodeRabbit

* **New Features**
  * Added configurable Sales Rules for quotes, sales orders, and invoices, including conditions, severity, item filters, assignments, and acknowledgment.
  * Added rule-violation prompts that support warnings and blocking errors.
  * Added sales-rule management screens, plan upgrade prompts, and notification recipient settings.
  * Added email notifications and direct links for sales-rule violations.
  * Unified storage and sales rules under a shared enforcement-rules experience.

* **Documentation**
  * Added Sales Rules reference documentation and expanded translations.

* **Bug Fixes**
  * Improved destination handling for order-derived and drop-shipped invoice lines during rule checks.


naveenkash:
## CodeRabbit findings — resolution summary (as of 7ade3ff9eb)

**Fixed in this push:**

- **direct-executor payload normalization** — both MCP backstops (line writes and the finalize/convert transitions) now resolve the payload exactly the way `executeFunction` does (named param key → single-key wrapper unwrap → flat args), so `{ args: {...} }` / `{ quotationLine: {...} }` requests can no longer skip evaluation and still reach the service.
- **`validationError(validation.error)`** — all four rule-save routes (sales + storage twins) now return the wrapped shape.
- **Flash headers discarded** — fixed in an earlier push (`3676e0ec62`): the four routes return `data({}, await flash(...))`.
- **Acknowledged evidence before the write** — the three `$lineId.details` routes now record acknowledged evidence only **after** the line write commits (blocked evidence still writes before the early return). The `.new` routes already did this.
- **`customerLocation` scoping** — `.eq("companyId", companyId)` added in `3676e0ec62`; now also switched to `.maybeSingle()` (missing location is an expected fail-closed state). `resolveSalesOrderShipTo`'s `salesOrderShipment` query was already companyId-scoped.
- **Digital-quote fail-closed** — the document evaluator already threw on lines-read errors; it now also throws on quote/salesRfq **header**-read errors (a null header would soften country rules).
- **Notification lookups** — the three `SalesRuleViolation` content queries are now scoped by `opts.companyId` (and return null without it).
- **Shipment-post lookup scoping** — the flagged read carries `companyId` on the current tree.

**Outdated (files/functions no longer exist after the enforcementRule merge):**

- `inventory.service.ts` / `items.service.ts` rule CRUD — replaced by the family-parametrized CRUD in `~/modules/shared`, which requires `companyId` on get/update/delete since `3676e0ec62`.
- `packages/ee/src/rules/item/server.ts` rules-query error discard — the current evaluator **throws** on rule/item/ship-to read errors.

**Declined, with reasoning:**

- **`bypassRls: true` on the convert routes** — `requirePermissions` already gates the action; the service-role client is used only for rule evaluation (the same pattern as quote finalize/convert across the branch), and `bypassRls: true` would *widen* the route's own client to service role rather than narrow anything.
- **Include `InApp` destination** — `notify.ts` adds `NotificationDestination.InApp` unconditionally ahead of the per-event defaults ("InApp is always added separately and cannot be opted out of"), so the email-only default already produces a topbar notification.
- **`item.customFields.*` never resolves (heavy lift)** — real, but pre-existing engine-wide behavior shared with the shipped storage family: neither evaluator loads subtype custom fields, the builder UI does not offer custom-field paths, and a hand-authored AST fails **closed** (required-field violation). Tracked as a follow-up rather than fixed inside this PR.

sidwebworks (approved):

sidwebworks:
Wait for the CI to pass - rest LGTM

PR #1354, opened 2026-08-08
docs(specs): RMA / customer returns module — spec and research

sidgaikwad:
## What

Design spec for a customer-returns (RMA) module, written to the `.ai/specs` convention — **no code in this PR**. Posting it as a PR so the design can be reviewed and commented on before any planning/implementation.

- Spec: `.ai/specs/2026-08-07-rma-module.md`
- Research: `.ai/research/2026-08-07-rma-module.md` (Business Central, NetSuite, Epicor, Infor, SAP B1, Odoo, Katana, Fulcrum — primary-source cited, gaps flagged)

## Design in one paragraph

A non-posting `salesReturnOrder` header + lines document in the Sales module — the model every document-first ERP surveyed converges on. It authorizes returns with optional links to the originating order/shipment/invoice (blind returns supported), receives goods through the **existing receipts system** by activating the pre-plumbed-but-unused `"Sales Return Order"` value in `receiptSourceDocument`, re-enters inventory at **original outbound cost** (exact cost reversing, with a per-reason zero-value hook for damaged goods), reactivates the **same `trackedEntity`** so serial/batch genealogy survives the round trip, escalates Scrap/Rework/MRB to the **existing quality Issues module** (Epicor's Fail→DMR pattern), and settles via the **existing AR credit `memo`** (plus a per-line credit-breakdown table) and an optional linked replacement sales order. Everything is additive — no renames, no drops, no new permission scopes.

## Review focus

All 14 open questions were resolved without a reviewer in the loop and are recorded inline in the spec's Open Questions section (marked **Autonomous:**) — they are decisions, not facts. Most worth scrutiny:

- Additive enum values: `disposition` gains `'Return to Customer'`; `nonConformanceAssociationType` gains `'Sales Return Order'`
- Additive columns: `memo.salesReturnOrderId`, `accountDefault.salesReturnsAccount` (+ a seeded contra-revenue account)
- Scrap/Rework GL routed through quality Issues (`post-nonconformance`) instead of RMA-native posting
- Exact-cost reversal from the shipment's cost layers — flagged as the riskiest part of the design (see Risks)

## Summary by CodeRabbit

* **Documentation**
  * Added research comparing return and RMA workflows across major business platforms.
  * Documented customer and supplier return processes, including authorization, receipt or shipment, reasons, dispositions, credits, replacements, repairs, and quality review.
  * Defined lifecycle states, validation rules, inventory and tracking behavior, costing, accounting outcomes, and completion safeguards.
[...]

PR #1392, opened 2026-08-14
feat: Returns module — customer RMAs and supplier returns

sidgaikwad:
Tracking spec: .ai/specs/2026-08-07-rma-module.md (merged via #1354)
Implementation plan: .ai/plans/2026-08-14-rma-module.md

## What this adds

The full Returns module v1 from the spec — both directions, everything additive:

**Customer RMAs (`salesReturnOrder`, Sales module)**
- Non-posting authorization document with `RMA000001` sequence ids, receipt-driven statuses (Draft → Confirmed → Partially Received → Received → Completed/Cancelled), per-line `closedComplete` short-close, and a customizable PDF generated + attached at Confirm.
- Lines link optionally to sales order / shipment / invoice lines; the "from document" picker offers only reversible quantities (shipped − already authorized), and the reversible cap is re-checked at Confirm inside a row-locked transaction so concurrent RMAs cannot jointly over-authorize a source line.
- Receiving reuses the receipts system end-to-end (`receiptSourceDocument 'Sales Return Order'` was pre-plumbed): `create` gains a `receiptFromSalesReturnOrder` case, `post-receipt` a return branch that re-enters stock at the **original outbound cost** (resolved from the source shipment's cost-ledger consumption rows, with current-cost fallback and a zero-value path for reasons flagged `inventoryValueZero`), posts Dr Inventory / Cr COGS, and **reactivates the same tracked entities** `On Hold` with a `Return Receipt` genealogy activity. Void reverses quantities, status, journal, entity states, and zeroes the receipt's own cost layers.
- Serial/batch return receipts pick **which expected entities arrived** (re-tagging the existing Consumed entity) instead of minting new serials.
- Dispositions use the existing quality vocabulary: Use As Is releases entities to Available; Return to Customer ships back through a `'Sales Return Order'` shipment; Scrap/Rework escalate into a pre-associated quality Issue (the Issue owns approvals and GL via the existing close flow).
- Credit generates the existing AR `memo` (header-level, unchanged shape) plus a `salesReturnOrderCreditLine` breakdown; `quantityCredited` derives from Posted memos so voiding un-credits automatically. `post-memo` routes return-linked memos to the new seeded **4900 Sales Returns** contra-revenue account (fallback `salesAccount`). Per-line restocking fee as a 0–1 fraction. Replacement orders draft a `resolvePrice`-priced sales order with a cross-link.

**Supplier returns (`purchaseReturnOrder`, Purchasing module)**
- Direction-flipped mirror: authorize → **ship** (via `shipmentSourceDocument 'Purchase Return Order'`) → credit. `RTS000001` ids, the supplier's own RMA number on the header, shared `returnReason` table, no disposition stage.
- Lines are always inventory-UOM (PO-line prices converted once at authoring); the picker is receipt-lineage-first with explicit blind-return flagging; the confirm cap bases on receipt lines → PO lines (× conversion factor) → invoice lines.
- `post-shipment` relieves at carried cost (Cr Inventory / Dr GR/IR Clearing, reversing the receipt posting), marks entities Consumed with a `Return Shipment` activity, and drives the Confirmed → Partially Shipped → Shipped ladder; void reverses it.
- Supplier credit memos post against GR/IR Clearing; replacement POs price from the linked PO line / supplier part.

**Quality bridge**
- Two new Issue association types (RMA lines, supplier return lines).
- An Issue with `'Return to Supplier'` dispositions gets a **Create Supplier Return** action: supplier auto-resolution (explicit → associated supplier → receipt-derived), per-quantity ownership rows on the association junction, idempotent re-invocation. `closeIssue` blocks while a linked return is open, writes off only the shipped-uncovered remainder (no double relief), and leaves return-shipped entities Consumed.

## Spec drift resolved during implementation (flagged, not silently redesigned)

- `nonConformanceAssociationType` is a TS const array, not a DB enum — the new keys are app-side; the junction tables follow the real NC-sibling shape (bare id PK, denormalized readable ids, quality_* RLS).
- `post-memo` derives `reasonAccount` at post time rather than reading it — resolved with one additive branch keyed on the new `memo.salesReturnOrderId`/`purchaseReturnOrderId` columns; the journal builder is untouched.
- `id('pro')` collides with `procedure` — supplier returns use `id('pret')`/`id('pretl')`.
- Shipments post `itemLedger.entryType 'Negative Adjmt.'`, not `'Sale'` — return identities ride on `documentType` (`'Sales Return Receipt'`/`'Purchase Return Shipment'`, both pre-existing unused values), matching the real machinery.
- The `documentSourceType` migration in this branch is a defensive no-op (`ADD VALUE IF NOT EXISTS` for values that already exist since 2024) — kept because it is already recorded in local migration history.

## Verification

- Scoped typechecks (`erp`, `@carbon/database`, `@carbon/documents`), Biome, and the full workspace test suite (23/23, including the conformance gate) pass; the new pure cost resolver has its own test file.
- Verified end-to-end against the local dev stack: RMA from a shipped serialized order (picker → cap rejection of over-authorization → confirm + PDF → receipt → posting at the exact original outbound cost 2 × 782.4183 vs 1899 sale price → same serials back On Hold with Return Receipt genealogy → Use As Is → Available → 10%-fee credit memo 3418.20 posted to 4900 Sales Returns → quantityCredited derived → Completed), receipt void (full rollback + zeroed layers), and the supplier mirror (RTS with supplier RMA number → confirm + PDF → shipment posting Cr Raw Materials / Dr GR/IR at carried cost → Shipped → AP credit against GR/IR Clearing).
- A step-by-step manual verification guide is at `.ai/docs/2026-08-14-returns-module-manual-checks.md`.

Out of scope per the spec review: creating customer RMAs from NCRs/quality issues (deferred to a follow-on spec).

## Summary by CodeRabbit

* **New Features**
  * Added customer RMAs and supplier return orders with lifecycle tracking, credits, replacements, and downloadable PDF documents.
  * Added returnable-item selection, serial/batch tracking, and return-specific receiving and shipping workflows.
  * Added configurable return reasons, quality-issue associations, supplier returns from issues, and inventory/accounting integration.
* **Bug Fixes**
  * Improved quality-issue closure and inventory handling for returned items.
* **Documentation**
  * Added manual test coverage and workflow guidance.
* **Localization**
  * Added translated return-related interface text across supported languages.


sidgaikwad:
Addressed the review in 49da06076. Summary of what changed and what was deliberately left:

**Fixed (CodeRabbit)**
- `create` edge fn: all three new return cases now reject re-sourcing when the existing receipt/shipment is not Draft, before any line delete.
- `closeIssue`: the linked-returns read and the tracked-entity picks read now fail the close on error instead of silently treating returns as absent.
- Shipment `lines.tracking`: the status-guarded update selects its rows and reports a conflict when zero rows matched.
- Receipt `lines.tracking`: `trackingType` union widened to include `returnEntity`.
- Bulk `update` routes (both modules): an unknown currency code now errors instead of falling through and leaving a stale `exchangeRate`.
- `$id.supplier-return`: explicit `supplierId` is validated against the company.
- Views: `companyId` predicates added to both list views' laterals; header FK indexes (`salesOrderId`, `replacement*Id`, `locationId`, PO equivalents) added; the 4900 backfill now sets `consolidatedRate: 'Average'` to match the seed.
- Credit modals: values interpolated inside single `<Trans>` elements.
- Line-details loaders: readable-id lookups collapsed to parallel embedded selects.
- Shipment/receipt `new` + shipment `details`: default-location guard and `getEdgeFunctionErrorMessage`/logging parity for the new return branches.
- Dead `salesReturnOrderComplete`/`purchaseReturnOrderComplete` path helpers removed; stale "follow-up" comments removed; returnable-lines and credit loaders log their query errors.
- The manual-check doc's destructive local-DB appendix replaced with rebuild-from-migrations guidance.

**Fixed (self-review, same commit)**
- FX: supplier-return pricing now reads `supplierUnitPrice` (supplier currency) rather than the base-currency `unitPrice` generated column — picker, issue bridge, and replacement PO.
- The supplier-return entity picker and bridge provenance check were anchored on a `Supplier` attribute nothing writes; both now resolve through the `Receipt` attribute → `receipt.supplierId`.
- Confirm caps: `companyId` predicates on the row-locked source reads (both modules).
- Line routes: membership of the line in the URL's order is verified, the parent id can't be swapped via the form, and quantities lock once the order leaves Draft (line adds are Draft-only) — so the confirm-time caps can't be bypassed after the fact.
- `post-shipment`: only a Posted shipment can void (a double void re-subtracted `quantityShipped`); `create` gates return-to-customer shipments on received statuses; `post-receipt` SRO posting allowlists statuses and its void refuses while a credit memo exists.
- Use-As-Is disposition release is one Kysely transaction (line + entity flips + genealogy rows).
- Cancel/complete guards fail on query errors; replacement creators gained a status guard, an error check, and a batched line insert; issue escalation checks its three NCR-link writes and requires `sales_update`.

**Left as-is, with reasons**
- Credit lines store full-precision values while the memo total is rounded at `currency.decimalPlaces` — per the numeric-precision standard (round at the settlement boundary only).
- `xid()` on `returnReason`/credit-line tables — matches recent child-table precedent (storage-type, cost-centers).
- `id('pret')` — `id('pro')` is already the procedure prefix; now commented at the column definition.
- `JSX.Element` in the documents block types — matches the sibling `salesOrder` block; the repo is on React 18 and the package typechecks.
- Tracked-entity pick setters use delete+insert without a transaction — picks are advisory pre-receipt data, re-saveable, and the authoritative validation happens at receiving.
- `lineNumber` read-then-write allocation and `bypassRls` on header create — both match the existing order-route conventions.
- Deno std 0.175.0 in the new test matches every other function test.
- `.in()` on `attributes ->> Receipt Line` verified equivalent to `.eq` against the live PostgREST.
- Returnable-lines pickers are bounded by supplier/customer + Posted + optional document filters; pagination there is deferred with the same shape as the app's other pickers.

**Known follow-ups (out of scope here)**
- Cross-link-type cap double counting: one line linked by `salesOrderLineId` and another by the same goods' `shipmentLineId` are capped independently. Closing this needs linkage-graph resolution across the three link columns.
- Return-shipment voids reverse the ledger/journal but don't restore consumed FIFO cost layers — the same gap the existing sales-order shipment void has; worth one shared fix.
- The GRNI offset on supplier-return credit memos books the memo amount while the shipment relieved carried cost; a variance line would make the account self-clearing.

sidgaikwad:
Addressed review rounds 2–4 in 4664b4242 + 845ce62d3. All 22 comments (20 inline + 2 body nitpicks) are handled:

**Fixed**
- Receive/Ship buttons disable while a create submission is in flight (`useNavigation`), on top of the server-side open-draft dedup.
- Line-details loaders: every readable-id lookup scoped by `companyId`, and the loader now verifies the line belongs to the URL's order (same check the action already had).
- Shipment `lines.tracking`: the guarded update runs **before** the stale-attribute clear — a status conflict no longer strips the previous assignment.
- The open-draft dedup lookups (receipt + both shipment cases) fail closed on query error.
- Returnable-lines modals: a line only counts as added when the action redirects to the new line's URL — validation errors and flashed failures no longer report success — and a mid-loop failure reports how many lines actually landed.
- Credit modals close on successful submit; the disposition select re-syncs to the loader's persisted value when its fetcher settles (reverts on failure).
- Blind return lines now price through `api/sales/resolve-price` (customer overrides, quantity breaks, pricing rules), falling back to `itemUnitSalePrice` on error; linked lines keep source-document pricing.
- Issue association route: both parent return-order lookups scoped by `companyId`; delete routes fail closed when the lines query errors.
- Migrations: composite `("id","companyId")` PK on both NC junction tables; the 4900 adoption UPDATE also requires `name = 'Sales Returns'` so a customized chart's unrelated 4900 is never adopted (NULL → documented `salesAccount` fallback).
- Accounting sync: `defaultEnabled: false` journal source types are now strictly opt-in at the posting gate — the return journals no longer auto-sync on upgraded integrations (the "always-on" set is unchanged for the frozen v2 types).
- PDF line blocks render "Total" without a literal `null` currency code (both directions).
- i18n: the three new returns strings are translated in all 12 locales, plus the four Russian wording fixes as suggested; the reason-load fallback now surfaces its error.

**Partial, with reasons**
- NC junction identity: the composite PK is applied, but a composite `("nonConformanceId","companyId")` FK is not possible — `nonConformance` has a bare-`id` primary key, as do all 11 existing `nonConformance*` association tables that reference it the same way. Changing that PK is out of scope for this PR; the app stamps `companyId` from the session on every insert (noted in the migration).
- Readable-id lookup errors in the details loaders degrade to "no linked document" rather than failing the page — the chip is informational, and throwing inside the parallel lookups would 500 an otherwise-valid line page for a cosmetic label.

aashu0148 (approved):

barbinbrad (approved):

PR #1401, opened 2026-08-16
feat(telemetry): capture work-done events server-side

BettrCallRaul:
## What does this PR do?

Records **work done on the platform** — jobs released, quantities reported, POs finalized — as PostHog events. Carbon currently measures screens opened, which cannot answer "is this account actually running production on Carbon?"

16 event types, 22 call sites across ERP and MES. No new dependencies, no migrations, no schema changes.

No linked issue — this comes out of the adoption-tracking work.

### Three design decisions worth reviewing

**No `posthog-node`.** Capture is a JSON POST with an api_key. The two things the SDK adds — batching and retry — are the wrong shape here: work events are low-frequency, and a retry without a stable id is how you double-count a released job. This follows the decision already made for Inngest in `packages/database/supabase/functions/lib/inngest.ts` and the shape of `packages/stripe/src/gtm-events.server.ts`.

**Deterministic event ids.** Every event carries a `uuid` derived from (companyId, event, recordId, discriminator) via RFC 4122 v5. PostHog's own de-duplication on that field is eventual and its docs say it is not guaranteed, so this id is what the warehouse de-duplicates on. Emitting the same occurrence twice is safe by construction, which is what lets every call site be fire-and-forget.

Creation-family events key on the row the work created (`productionEventId`, `productionQuantityId`) so repeat work in a shift stays distinct. State transitions key on the document, so a re-post collapses.

**Same gate as the browser SDK.** Emits only when `POSTHOG_PROJECT_PUBLIC_KEY` is set and `CONTROLLED_ENVIRONMENT` is off — `apps/erp/app/entry.client.tsx:31`. Server capture must not be reachable anywhere the browser's is not, or an ITAR deployment that believes analytics are off would start shipping business records.

Deliberately **not** named `.server`: MES service files are in the client graph (React components import their types), so a `.server` specifier fails the React Router build. That is also why the hash is plain TypeScript rather than `node:crypto`.

### The events

| Module | Events |
|---|---|
| Production | `job_created`, `job_released`, `job_operation_started`, `production_quantity_reported`, `job_operation_finished`, `job_completed`, `scrap_reported` |
| Inventory | `receipt_posted`, `shipment_posted`, `picking_list_completed` |
| Sales | `quote_sent`, `quote_accepted`, `sales_order_confirmed` |
| Purchasing | `purchase_order_finalized` |
| Invoicing | `sales_invoice_posted`, `purchase_invoice_posted` |

Payloads carry ids, enums, counts and quantities only — never money, part numbers, names or free text. Each also carries `companyId`, the actor as `distinct_id`, `module`, `work_event: true`, and a `company` group.

Several properties exist to stop a number being read the wrong way:

- `job_created.source` (erp / bulk / mrp / salesOrder / kanban) separates human intent from MRP output.
- `production_quantity_reported.source` (mes / mes_qr / erp) is the best single measure of whether the shop floor app is genuinely in use or a clerk is typing yesterday's paper traveller into the back office.
- `purchase_order_finalized.approvalRequired` — a PO stopped at an approval gate is not money committed.
- `job_released.priorStatus` — re-saving an already-released job never emits.

The six production events share a `jobId`, so created → released → started → reported → finished → completed joins into one chain per job.

## Visual Demo

Not applicable — no UI. Verifiable output is the event stream itself; see below for how to watch it locally.

## How should this be tested?

**Automated** — 28 unit tests, `pnpm --filter @carbon/lib test`. Full suite 24/24 packages, build 8/8, both re-run after rebasing onto `c967003de`.

They cover that a retry produces the same uuid and two different jobs do not; that a reverse-proxied host keeps its base path; that all three off-switches actually silence it; that a PostHog rejection, a network failure and a fire-and-forget call each resolve instead of throwing. Plus SHA-1 against the standard vectors and the RFC 4122 §A worked example, so the hash is a genuine v5.

**What is not covered: the 22 call sites.** No test proves that releasing a job calls `trackWorkEvent`, or that the call sits below the rollback catch rather than above it. Placement was verified by reading each seam, which is weaker than a test. This is the main thing to look at in review.

**Locally**, capture is off by default because `POSTHOG_PROJECT_PUBLIC_KEY` is unset. To watch events without pointing dev traffic at the real project, run any local HTTP sink and set:

```
POSTHOG_API_HOST="http://localhost:8047"
POSTHOG_PROJECT_PUBLIC_KEY="phc_local_dev_sink"
```

Then release a job, or on MES start an operation → report a quantity → finish it.

**Before merging**, one environment check that is cheap and is the single largest risk: confirm `POSTHOG_API_HOST` on the deployed ERP, and that the browser is currently ingesting through it. The emitter POSTs to `${POSTHOG_API_HOST}/e/` — deliberately the same host and path `posthog-js` uses, because that is the one combination proven against the deployed value rather than against the docs. Capture never throws, so a wrong host loses every event with no signal beyond a log line.

`packages/lib/src/telemetry/VERIFICATION.md` is the post-deploy runbook: what to check at one hour, one day and three days, with the queries and the pass criteria. The T+3 check reconciles counts against `productionQuantity`, `productionEvent` and `itemLedger` — append-only tables, rather than status columns, which would be circular.

## Known gaps

Documented in `packages/lib/src/telemetry/README.md` rather than left to be discovered as zeroes:

- **`scrap_reported` from the MES floor** and the **serial/batch branches of `production_quantity_reported`** write inside the `issue` Deno function, which never returns the inserted row id. There is no key that makes one posting distinct, and keying on the operation would under-count a shift's repeats. Fix is to return the id from that function.
- **`quote_accepted` from the customer portal** is anonymous — `actorId` is deliberately null for a customer's own acceptance. This is correct, not a gap.

## Open question for review

`adoption-tracking/spec.md` scoped this as an external service over the audit API with zero product code. This is product code — the lightest version I could make it, but the tradeoff is real and worth a decision rather than an assumption.

There is also no alerting sink anywhere in the repo: `trackWorkEvent` is fire-and-forget, and PostHog drops events past a billing limit **while still returning HTTP 200**. Total silent loss is possible and the only trace is a log line. The T+3 reconciliation is the mitigation, not a fix.

## Summary by CodeRabbit

* **New Features**
  * Added work-event tracking across production, sales, purchasing, inventory, invoicing, shipping, and warehouse workflows.
  * Captures key lifecycle actions such as job creation and completion, quote activity, order confirmation, quantity reporting, picking-list updates, and document posting.
  * Records workflow sources and relevant context to improve activity visibility and reporting.
* **Documentation**
  * Added guidance for telemetry setup, verification, event coverage, and monitoring.
* **Tests**
  * Added coverage for event delivery, attribution, deduplication, configuration, and failure handling.


apps/erp/app/routes/x+/shipment+/$shipmentId.post.tsx
```diff
@@ -191,6 +192,9 @@ export async function action({ request, params }: ActionFunctionArgs) {
     );
   }
 
+  /** Set by the catch below when the post was rolled back to Draft. */
+  let reverted = false;
```
barbinbrad, line 196: I don't think this is necessary. We just want to know whether they're using it. Whether it works is a different problem.

PR #1409, opened 2026-08-17
fix(production): batch N+1 method-tree queries in get-method and recalculate

naveenkash:
## What does this PR do?

Fixes job creation taking minutes (and sometimes producing partial job data) for items with large BOMs. Both `get-method` (`itemToJob`) and `recalculate` (`updateJobQuantities`) walked the method tree issuing 2–4 queries per node on a single transaction connection — ~600–800 sequential round trips for a 32-assembly / 44-operation / 259-material BOM — which exceeded the Supabase client's 25s per-attempt timeout and caused `fetchWithRetry` to stack three concurrent delete-and-recreate runs.

Two commits:

1. **`fix(production)`** — prefetch each lookup table once for the whole tree (`itemReplenishment`, `methodOperation` + embeds, default storage units), compute quantities in memory with unchanged math, write set-based via `UPDATE … FROM (VALUES …)`; remove leftover per-node debug logging. Measured on a local stack: `get-method` 25s+ (timing out) → ~4.5s, `recalculate` 27.5s → ~1.1s, with byte-identical results (md5 checksums over all recomputed quantity columns match the old implementation's output).
2. **`refactor(production)`** — extract recalculate's quantity cascade into a pure, I/O-free `lib/job-quantities-engine.ts` following the `mrp-engine.ts` pattern, with cycle detection that reports corrupt/cyclic tree data (`cycleNodeIds`) instead of recursing forever, plus `deno test` coverage; `get-method`'s prefetch walk gets the same cycle guard.

N/A — backend performance fix; before/after timings above serve as the demonstration.

- [x] I confirm automated tests are in place that prove my fix is effective or that my feature works. (`lib/job-quantities-engine.test.ts`, 6 deno tests pinning the cascade math, scrap fallback semantics, and cycle handling; the get-method traversal has no unit harness and was verified by live invocation with output-equivalence checksums.)

## How should this be tested?

- Seed or use an item whose make method has a deep/wide BOM (tens of sub-assemblies, hundreds of materials).
- Create a job for that item: it should complete in seconds, with a single `get-method` and a single `recalculate` invocation in the edge-runtime logs (no 25s retry stacking), and the job's make methods / operations / materials should match the source method counts.
- Re-run `recalculate` (`jobRequirements`) on an existing job and confirm `jobMaterial.scrapQuantity`/`estimatedQuantity` and `jobOperation.targetQuantity`/`operationQuantity` values are unchanged from the previous implementation.
- Unit tests: `cd packages/database/supabase/functions && deno test lib/job-quantities-engine.test.ts`.

### DIFF
--- a/.ai/lessons.md
+++ b/.ai/lessons.md
@@ -1023,3 +1023,13 @@ canvas hosting Radix popovers/selects.
 **Rule:** Programmatic submits inside a `ValidatedForm` must pass a submitter: `form.requestSubmit(form.querySelector('button[type="submit"]'))`, which means the form needs a real submit button (good for accessibility anyway). Never use `form.submit()` in a React Router app. When a form renders errors from `fetcher.data`, verify the submit path actually reaches the fetcher — an unreachable error branch looks identical to "no errors happen".
 
 **Applies to:** `packages/form/src/components/InputOTP.tsx`, `packages/form/src/ValidatedForm.tsx`, any auto-submitting form field.
+
+## Batched PostgREST `.in()` with hundreds of ids blows the gateway URL limit — use Kysely for big id-list reads in edge functions
+
+**Context:** Fixing the N+1 traversal in `get-method`'s `itemToJob` by prefetching `itemReplenishment` for a whole method tree (260 item ids) with `client.from(...).in("itemId", ids)` chunked at 200 ids per request.
+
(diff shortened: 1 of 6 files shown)


sidwebworks:
Let's hold on merging this one - I need to make some changes

sidwebworks (approved):
Let's merge it for now and I'll do a follow up refactor PR

PR #1419, opened 2026-08-18
Show change notices under Used In and on the issue sidebar

aashu0148:
## What does this PR do?

- Adds a "Change Notices" group to the Used In tab on part and tool pages, listing every change notice the item appears on with its current stage. Previously the page only hinted that a change notice existed, and only while that notice was still open, so there was no way to see the item's change history from the item itself.
- Moves the change notice list on an issue into the left sidebar alongside the other things an issue is linked to, instead of tucking it at the bottom of the right properties panel. It is read-only there (you can open a change notice, but you cannot attach one from the issue), which matches how change notices actually get linked.
- Fixes long status labels getting clipped in both of those lists, and removes the hover "Open" button from change notice rows on the part page since clicking the row already opens it.
- Replaces four hand-written copies of the "is this change notice still open?" rule with one shared check, so the item alert, the revision lock, and both item pages can no longer disagree.

## Images 
image-0001
image-0001

## Summary by CodeRabbit

- **New Features**
  - Added Change Notices to Part and Tool “Used In” trees with status badges, counts, and navigation links.
  - Added a read-only Change Notices section to Issue associations.
  - Displayed Change Notice statuses consistently across item and issue views.
- **Bug Fixes**
  - Improved open Change Notice detection for item locking and alerts.
  - Prevented new versions and revisions when Change Notice data cannot be loaded, with a reload prompt.
  - Preserved existing permission controls and locking behavior while showing all notice statuses.

### DIFF
--- a/.ai/plans/2026-08-18-change-notices-used-in-and-issue-sidebar.md
+++ b/.ai/plans/2026-08-18-change-notices-used-in-and-issue-sidebar.md
@@ -1,0 +1,732 @@
+# Change Notices as a "where referenced" category — implementation plan
+
+**Spec / source:** `.ai/specs/2026-08-18-change-notices-used-in-and-issue-sidebar.md`
+**Branch:** `feat/cn-improvements`
+
+No migrations, no schema changes, no new service functions. Every query already
+exists and is already called by the routes being changed.
+
+## Progress
+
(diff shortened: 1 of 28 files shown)


naveenkash (approved):
Lgtm

PR #1425, opened 2026-08-18
feat(jobs,auth): auto-disable inactive accounts in cleanup job (NIST 3.5.6)

barbinbrad:
## What & why

NIST 800-171 **3.5.6 / AC-2(3)** — no scheduled job currently deactivates idle accounts. This adds a new `disable-inactive-accounts` step to the **existing** daily cleanup Inngest job (`packages/jobs/src/inngest/functions/scheduled/cleanup.ts`) — no new cron, additive only.

In a controlled (CUI/ITAR) environment **only** (`CONTROLLED_ENVIRONMENT`), the step:
1. Reads active employee memberships + last-login from GoTrue `auth.audit_log_entries` (action `login`) via a raw Kysely query (auth schema isn't exposed to PostgREST).
2. Selects accounts idle beyond **35 days** (NIST AC-2(3) baseline) using a pure, unit-tested helper.
3. Deactivates each via the existing `deactivateUser` flow.
4. Scrubs residual `userId` references the flow leaves behind, per `.claude/rules/user-employee-job-relationships.md`: **assignee columns** (33 tables), **`companySettings` notification-group arrays** (11 columns), and **`workCenterEmployee`**.
5. Emits a structured `account_auto_deactivated` auth event via `logAuthEvent` for the SIEM trail.

Non-CUI deployments are a no-op. Cron schedule, existing steps, and deactivation semantics are unchanged.

## Key design
- **Pure selection helper** `selectInactiveAccounts` — TDD'd (10 vitest cases). Threshold math via `@internationalized/date` (no JS `Date`); `createdAt` floor so never-logged-in accounts aren't disabled the moment they're created; `protected` (console/admin/developer) + system-id exclusions; safety guards return nothing on bad config / bad data.
- **Per-run cap (200)** drains a first-pass backlog over successive days (mirrors the agent-thread step's `limit(200)`).
- Idempotent on Inngest retry — the candidate query excludes already-disabled employees; re-scrub is a no-op.

## Gates
- ✅ `turbo typecheck --filter=@carbon/jobs --filter=@carbon/auth`
- ✅ `pnpm --filter @carbon/jobs test` — 486/486 (incl. 10 new)
- ✅ `pnpm run lint` — 33/33 tasks (new files clean)

## UNVERIFIED (no runnable stack/DB in this environment)
- Live cron execution of the step.
- The raw `auth.audit_log_entries` query against real GoTrue data (written from GoTrue's standard schema; not executed here).
- End-to-end deactivation + scrub against a real tenant.

## Coordination note
`packages/auth/src/services/auth-events.server.ts` (`logAuthEvent`) existed only in the base branch's **uncommitted** work; it's vendored here (self-contained, depends only on `@carbon/logger`) so this branch compiles/tests in isolation, adding the `account_auto_deactivated` event type. If the base commits its copy first, expect a trivial union-type merge (keep both).

### DIFF
--- a/.ai/plans/2026-08-18-nist-2.6-inactive-account-disable.md
+++ b/.ai/plans/2026-08-18-nist-2.6-inactive-account-disable.md
@@ -1,0 +1,54 @@
+# Plan: NIST 2.6 — Auto-disable inactive accounts in the cleanup job (3.5.6)
+
+## Goal
+Add ONE new `step.run` to `packages/jobs/src/inngest/functions/scheduled/cleanup.ts`
+that deactivates employee accounts idle beyond a threshold, in a controlled (CUI)
+environment. Reuse `deactivateUser`/`deactivateEmployee`. Determine last-login from
+`auth.audit_log_entries`. Scrub residual `userId` references. Emit a structured auth
+event per deactivation. Selection logic is a pure, TDD'd helper.
+
+## Constraints
+- ADDITIVE. No change to cron schedule, existing steps, or deactivation semantics.
+- No JS `Date` for age math — `@internationalized/date` + `@carbon/utils` `datetime.timestamp()`.
+- Gate the whole step on `CONTROLLED_ENVIRONMENT` (non-CUI deployments unaffected).
+- No execution against real DB; selection helper proven by vitest (red→green).
+
+## Tasks
+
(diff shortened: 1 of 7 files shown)


barbinbrad:
We can't enable this until we have a some login logs built up

PR #1453, opened 2026-08-21
Fix/timezones picker

aashu0148:
## What does this PR do?

- Timezone options are now a single line each, like `America/New York (GMT-04:00)`, instead of a title with a second line of detail underneath (the two-line rows made the list hard to scan).
- Searching by name works: typing "eastern time", "pacific", "EST" or "EDT" now finds the right zones. Previously only the raw technical id was searchable, so those searches returned nothing.
- Searching with spaces instead of slashes works: "asia kolkata" now finds `Asia/Kolkata`. The slash used to block the match, which is what made the picker feel broken.
- Results are ranked better: a whole-word match wins over a partial one, so searching "EST" leads with actual Eastern zones instead of Creston and Bucharest.

## Summary by CodeRabbit

* **New Features**
  * Improved timezone labels with readable display names and GMT offsets.
  * Enhanced timezone search to recognize alternate separators, display names, abbreviations, and related keywords.
  * Added flexible combobox filtering with support for custom filter behavior and searchable keywords.
  * Improved result ranking across labels, helper text, keywords, and combined text.
  * Exposed reusable combobox filtering utilities and option types for custom integrations.


packages/react/src/Combobox.tsx
```diff
@@ -222,23 +233,36 @@ function VirtualizedCommand({
...
+        [option.helper, option.keywords].filter(Boolean).join(" ")
+      );
+      const hiddenIndex = hiddenText.indexOf(query);
+      if (hiddenIndex !== -1 && wordStartAt(hiddenText, hiddenIndex)) {
+        scored.push({ option, score: 200000 });
+      } else if (labelIndex !== -1) {
+        scored.push({ option, score: 300000 + labelIndex });
+      } else if (hiddenIndex !== -1) {
+        scored.push({ option, score: 400000 });
+      }
```
sidwebworks: Why are we altering the behaviour for all the comboboxes?

Also the old code matched against ``${label} ${helper}`` so something like `PART-001 Suffix` worked but now doesn't give any results.

packages/database/supabase/functions/lib/types.ts
```diff
@@ -34473,6 +34473,443 @@ export type Database = {
           },
         ]
       }
+      onshapeItemSyncState: {
```
sidwebworks: Unrelated types change - let's revert this.

packages/form/src/components/Timezone.tsx
```diff
@@ -17,29 +18,35 @@ type TimezoneProps = Omit<ComboboxProps, "options"> & {
...
+            keywords: [
+              option.value,
+              getTimezoneDisplayName(option.value),
+              abbreviations
+            ]
+              .filter(Boolean)
+              .join(" ")
+          };
+        })
+        .sort((a, b) => a.label.localeCompare(b.label))
```
sidwebworks: The `.sort()` should happen on the result of the `flatMap` so it's actually A-Z in the end.

packages/react/src/Combobox.tsx
```diff
@@ -209,6 +211,15 @@ type VirtualizedCommandProps = {
   setOpen: (open: boolean) => void;
 };
 
+/** Treats `/`, `_` and `-` as spaces so "asia kolkata" matches "Asia/Kolkata". */
+function normalizeSearchText(text: string): string {
+  return text
+    .toLowerCase()
+    .replace(/[/_-]+/g, " ")
```
sidwebworks: This means strings can never have "-" sign? This will change a timezone offset from `-3:30` to `3:30`

sidwebworks (changes requested):
I think we should handle searching differently. A combobox shouldn't try to provide a one size fits all searching algorithm rather give an escape-hatch to allow search functionality to be hooked into via composition.

So even something like [fuse.js ](https://www.fusejs.io/) can be used where we want to use it.

packages/react/src/Combobox.tsx
aashu0148: not really, its not going to affect what is there in the list. 
if you search 3:30 it will give you both -3:30 and +3:30

sidwebworks (approved):
LGTM - great work!

PR #1472, opened 2026-08-24
feat(documents): itemize add-on prices on quote/order/invoice PDFs

sidgaikwad:
## Problem

On the quote PDF, add-ons ("Delivery & Installation", "Design & Engineering", etc.) were listed by **name only** under **Tax & Fees** — the amounts appeared solely in the aggregate **Fees** total, so a reader could not tell what any individual add-on cost. The sales order and invoice PDFs had the same gap: they listed the fee categories with no numbers next to them.

## Change

Each entry in the per-line **Tax & Fees** breakdown now shows its amount, right-aligned within the description column (matching the in-app `QuoteSummary` presentation).

- **Quote** — every named additional charge (`quoteLine.additionalCharges`) plus Shipping and Tax, using the currency-converted per-quantity amount. Taxable vs non-taxable charges are respected in the tax line, exactly as the totals already computed them.
- **Sales order & invoice** — the taxable **Add-On**, **Non-Taxable Add-On**, **Shipping**, and **Tax** amounts.

This is a presentation-only change in `@carbon/documents` — **no schema, migration, or conversion changes**.

### Note on the sales order / invoice breakdown

Only quotes store the itemized, named charge map. When a quote is converted to a sales order, the individual charges are collapsed into two aggregate fields (taxable `addOnCost` + `nonTaxableAddOnCost`) and the names are not retained; order → invoice copies those two fields through. So orders and invoices show the fee **categories** with amounts rather than the named charges. Carrying the named breakdown all the way through would require a schema + conversion change and is intentionally out of scope here.

## Before / After

**Quote** — each add-on now shows its price:

Quote before and after

**Sales order:**

Sales order before and after

**Invoice:**

Invoice before and after

## Testing

- Rendered the quote, sales order, and sales invoice PDFs with representative add-on data (three named charges totalling $2,150, plus shipping and a non-taxable charge) and confirmed each amount renders correctly and the tax line reflects only taxable charges.
- `pnpm --filter @carbon/documents typecheck` ✅
- `pnpm --filter @carbon/documents test` ✅ (47 passing)
- `biome check` ✅

## Summary by CodeRabbit

* **New Features**
  * Added Supplier columns to parts, materials, tools, consumables, and services tables.
  * Supplier avatars, additional-supplier tooltips, filtering, and supplier-name exports are now available.
* **Bug Fixes**
  * Improved supplier data availability and filtering across item lists.
  * PDF documents now display amounts for shipping, add-ons, non-taxable add-ons, and taxes alongside their labels.
  * Quote charges now consistently show formatted, currency-converted amounts where applicable.

### DIFF
--- a/packages/documents/src/pdf/blocks/LineItemsBlock.tsx
+++ b/packages/documents/src/pdf/blocks/LineItemsBlock.tsx
@@ -114,24 +114,52 @@ export function LineItemsBlock({
                       Tax & Fees
                     </Text>
                     {lineShippingCost > 0 && (
-                      <Text style={tw("text-[9px] text-gray-600")}>
-                        - Shipping
-                      </Text>
+                      <View style={tw("flex flex-row justify-between")}>
+                        <Text
+                          style={tw("text-[9px] text-gray-600 flex-1 pr-2")}
+                        >
+                          - Shipping
+                        </Text>
+                        <Text style={tw("text-[9px] text-gray-600")}>
+                          {numberFormatter.format(lineShippingCost)}
+                        </Text>
+                      </View>
                     )}
                     {lineAddOnCost > 0 && (
-                      <Text style={tw("text-[9px] text-gray-600")}>
-                        - Add-On
-                      </Text>
+                      <View style={tw("flex flex-row justify-between")}>
+                        <Text
+                          style={tw("text-[9px] text-gray-600 flex-1 pr-2")}
+                        >
+                          - Add-On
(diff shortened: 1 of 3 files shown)


barbinbrad (approved):

barbinbrad (changes requested):
it looks like the changes from the other branch leaked into this one. you can just fix this branch, and add the supplier avatar group component here

barbinbrad (approved):

PR #1480, opened 2026-08-25
fix(scripts): never skip the email scrub when the restore drops the connection

naveenkash:
## Problem

`SCRUB_EMAILS=1` appeared to do nothing: after a restore, real production email addresses were still in the local DB, with no error on screen.

Root cause (reproduced from a real run's `/tmp/restore-errors.log`): the plain-text restore path pipes the dump into psql with **no failure guard**:

```bash
sed -E ... "$BACKUP_FILE" | $PSQL_SA -v ON_ERROR_STOP=0 2> /tmp/restore-errors.log
```

When the dump comes from a newer Supabase than the local stack, a `COPY` into a schema-drifted table (e.g. `auth.custom_oauth_providers` with a column the local GoTrue lacks) fails as a statement, psql loses sync and interprets the raw COPY data as commands (`invalid command \N` × hundreds), and the server connection dies. psql then exits 2 — and under `set -euo pipefail` the **whole script dies right there**, silently: no localization, no admin grant, and **no email scrub**, with the already-restored prod data (and its real emails) left on disk. All the evidence goes to the log file, not the terminal. The `pg_restore` branch already had `|| true`; the psql branch didn't.

## Fix

- Guard the psql pipeline with `|| true` (matching the pg_restore branch), then detect a dropped connection in the error log, print a loud "restore likely incomplete" warning, and **wait for Postgres to come back** (Docker restarts it) so the post-restore steps still run against a live connection — or fail loudly if it never returns.
- Make the scrub verification a real gate: when `SCRUB_EMAILS` is set and any non-admin email survives, the script **exits 1** instead of printing counts and ending with ✅.
- Add the two email columns the scrub list was missing: `companyAccountsPayableBillingAddress.email` and `companyAccountsReceivableBillingAddress.email` (guarded by the existing `information_schema` existence check, so schemas without them are unaffected).

## Verification

- `bash -n` passes.
- Ran the scrub + verification blocks verbatim against a local dev DB with synthetic rows: emails rewrite to `u_<hash>@example.test`, clean verification passes (exit 0), a deliberately leaked address makes verification exit 1.
- Wait loop: instant pass against a healthy server; correctly exits 1 with the loud message when the server never comes back.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database restoration reliability by waiting for the database service to recover before continuing.
  * Restores now distinguish unreadable backups, connection failures, and restore errors, and fail clearly when recovery is unsuccessful.
  * Enhanced email scrubbing for billing-address records, including accounts-payable and accounts-receivable data.
  * Added comprehensive verification to detect missing tables, failed scrubbing, and email data exceeding permitted administrative records.

### DIFF
--- a/scripts/restore-database.sh
+++ b/scripts/restore-database.sh
@@ -53,6 +53,10 @@
 #   - Refuses to run if the worktree isn't registered in ~/.carbon/dev-ports.json.
 #
 set -euo pipefail
+# Private, unpredictable error-log path (a fixed /tmp name is symlink-attackable
+# and can be pre-created by another local user).
+RESTORE_LOG="$(mktemp "${TMPDIR:-/tmp}/restore-errors.XXXXXX.log")"
+RESTORE_INCOMPLETE=""
 ADMIN_EMAIL="${ADMIN_EMAIL:-}"
 ADMIN_PASSWORD="${ADMIN_PASSWORD:-localpass}"
 SCRUB_EMAILS="${SCRUB_EMAILS:-}"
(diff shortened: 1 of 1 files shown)


naveenkash:
Addressed all three CodeRabbit findings in 175d07846:

1. **Pipeline failure suppression (Major)** — replaced the blanket `|| true` with `PIPESTATUS` capture. An unreadable backup or a psql that never connected now fails loudly before any post-restore mutation; only a detected mid-restore connection drop continues (so the email scrub still runs over whatever landed), and the script then exits nonzero at the end so callers see the restore as failed.
2. **Verification missing billing-address columns (Major)** — the LEAKED assertion is now a mapping-driven DO block covering every scrubbed (table, column) pair, guarded by information_schema so dumps predating a table pass instead of erroring; an empty verification result is itself a failure. Cross-checked against a restored dataset: the mapping-driven total matches a manual per-table sum exactly.
3. **Predictable /tmp path (Minor)** — the error log is now created with `mktemp`.

All failure paths exercised: unreadable backup → exit 1, unreachable server → exit 1, connection-loss marker → continues with warning and final nonzero exit, clean restore → exit 0.

naveencarbon (approved):

PR #1490, opened 2026-08-27
Run scheduled MRP without billing rows; keep hand-typed quote material costs

sidwebworks:
Two fixes.

**MRP scheduling** — the every-3-hours scheduler built its work list from `companyPlan`, which is only written by Stripe checkout and seeded nowhere. Self-hosted, community and local-dev installs therefore had an empty work list and silently never ran MRP, with a green Inngest run to show for it. It now enumerates `company`; on Cloud, companies with a `Canceled` subscription are skipped (the weekly job deletes those). The selection rule is a pure function with unit tests, and the scheduler warns when the list comes back empty. (`weekly.ts` was checked too — it already enumerates `company` and its deletion pass is Cloud-gated.)

**Quote material costs** — every quote method mutation re-derived each "Purchase to Order" material's unit cost from supplier price breaks, so a cost typed in the BOM editor was saved and immediately overwritten in the same request (only where the item had a `supplierPart` row). `quoteMaterial.unitCostSource` (`system`/`manual`) now records the cost's origin, and a manual cost wins everywhere a buy cost is resolved via the shared `resolveBuyUnitCost` (app + edge runtime). Existing rows are deliberately not backfilled — the overwrite already ran on them, so marking them manual would freeze a supplier-derived number. The migration recreates the `quoteMaterialWithMakeMethodId` view (a view's `SELECT *` is frozen at creation) and the two `get_quote_methods` readers.

**Verification**
- `pnpm --filter @carbon/jobs test` — 496 passed, incl. 6 new `selectCompaniesForMrp` tests
- `turbo typecheck --filter=erp --filter=@carbon/jobs` — clean
- `pnpm run lint` — clean
- Migration applied on local dev DB; column present on table, view, and both functions

Note: generated types/swagger were updated for the new column only; a stray local schema (in-flight batching branch) was kept out of the regen.

## Summary by CodeRabbit

* **New Features**
  * Quote materials now distinguish between system-calculated and manually entered unit costs.
  * Manually entered costs remain unchanged when quantities or supplier pricing change.
  * Material copies and quote revisions preserve cost-source information.

* **Bug Fixes**
  * Scheduled MRP runs now include eligible companies even when plan records are unavailable.
  * Cancelled Cloud subscriptions are excluded from planning, while other companies continue to be processed.
  * Improved MRP reliability for large company lists and data-fetch failures.
  * Added validation to prevent missing or invalid cost-source values.

### DIFF
--- a/.claude/rules/mrp-system.md
+++ b/.claude/rules/mrp-system.md
@@ -18,10 +18,22 @@ Deno edge function** — NOT Trigger.dev, and not inline in the app server.
 
 1. **Scheduled job** — `packages/jobs/src/inngest/functions/scheduled/mrp.ts`.
    `inngest.createFunction({ id: "mrp", retries: 2 }, { cron: "0 */3 * * *" }, …)`
-   — every 3 hours. Fans out **per company**: selects all rows from `companyPlan`
+   — every 3 hours. Fans out **per company**: selects all rows from `company`
    and, for each, calls `serviceRole.functions.invoke("mrp", { body: { type:
    "company", id, companyId, userId: "system" } })`. There is no location-scoped
    cron — only company-wide.
+
+   It enumerated `companyPlan` until 2026-08-26. MRP is not in `FEATURE_PLANS`,
+   so that was never a billing gate — just a convenient list of companies — but
+   the table is only written by Stripe checkout and is seeded nowhere, so every
+   self-hosted, community and local-dev install had an empty work list and
+   silently never ran MRP, reporting a green Inngest run. Do not reintroduce it:
+   the work list is `company`, and a company with no plan row must still run.
+   On **Cloud only**, companies whose `stripeSubscriptionStatus` is `'Canceled'`
+   are skipped, because `weekly.ts` deletes those. The selection rule is the pure
+   `selectCompaniesForMrp` (`scheduled/mrp-companies.ts`), unit-tested in its
+   sibling `.test.ts`; the scheduler logs a `warn` when the list comes back
+   empty, so "no work" can never again look like "worked fine".
 
 2. **Manual trigger** — POST `apps/erp/app/routes/api+/mrp.ts` (permission
    `update: "inventory"`). Reads `?location` query param; calls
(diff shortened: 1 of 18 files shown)


sidwebworks:
Addressed in 04472d65d6 — three of the five taken, two skipped with reasons.

**Fixed**

- **Paginate both MRP reads** — real, and the same bug class as the PR itself. `max_rows = 1000` is set in `config.toml`, and the sibling `weekly.ts` already pages for exactly this reason. Both reads now use `fetchAllFromTable` with a stable `.order("id")` (paging without an ORDER BY is not deterministic). The dev stack doesn't enforce the cap, so a dropped tail would have been invisible locally.
- **Throw on company-enumeration failure** — real. Returning made the step succeed having planned for nobody and never spent the two configured retries. A failed `companyPlan` read still doesn't throw: it leaves `plans` null and plans for everyone, which is the fail-safe direction.
- **Manual-cost race in both quote forms** — real, and a genuine miss. The `unitCostSource` guard read the value captured when the callback was built, so a cost typed while the `itemCost` and price-break lookups were in flight got overwritten when they resolved. The functional updater now re-checks against current state. Same bug the PR is about, reached a different way.

**Skipped**

- **`enum` on `unitCostSource` in `swagger-docs-schema.ts`** — that file is generated: `pnpm generate:swagger` dumps PostgREST's own OpenAPI output verbatim, so a hand-added enum is reverted on the next regen (and AGENTS.md forbids editing generated schema). PostgREST emits `enum` only for real Postgres enum types; this column is `TEXT` + CHECK, matching the sibling `quoteLinePrice.priceSource` from July, which appears in the same file as a bare `type: "string"`. The generated output is correct for the actual schema.
- **`priceMap` validation in the MCP dispatch path** — the throw is real (`priceMap[itemId]` on null), but it's pre-existing and not introduced here: `shared_lookupBuyPriceFromMap` is already exposed with the identical untyped `priceMap: {}` schema. The root cause is that `scripts/generate-mcp.ts` sweeps pure exported helpers out of `*.service.ts` into the tool surface — `resolveBuyUnitCost` is classified `WRITE` with `companyId` injected, despite taking no client and doing no I/O. Worth fixing at the generator rather than patching one call site, and there's no tenancy or write impact (pure function). Filed separately.

Verification: `@carbon/jobs` tests 496 passed, typecheck clean on `erp` + `@carbon/jobs`, lint clean.

PR #1530, opened 2026-08-31
fix(accounting): take the refresh rate from the document, not the request

BettrCallRaul:
> Stacked on #1526 — it uses `resolveCurrencyAndRate`, introduced there. Base will need retargeting to `main` once #1526 merges.

## Problem

The six `*.exchange-rate` routes read `currencyCode` from the posted form, look up **that** currency's rate, and write only `exchangeRate` — never `currencyCode`. A crafted POST can therefore stamp one currency's rate onto a document that keeps its own code.

Posting `SEK` to a USD invoice left it denominated in USD at SEK's rate of 11.24, which mispriced every line through the `converted*` generated columns. Found while verifying #1527.

The UI always posts the document's own currency (e.g. `QuoteProperties.tsx:438-447`), so this behaves as a refresh in practice. That is what makes it worth closing rather than living with — the parameter carries no information the server does not already have.

## Fix

Every one of these routes already loads its document for the locked check, so the currency is read from there and the form is no longer parsed at all. The lookup goes through `resolveCurrencyAndRate`, so a refresh cannot install a zero or missing rate either.

## Verification

Typecheck and lint clean. The client-supplied value is now ignored by construction — no route calls `request.formData()` any more.

Runtime re-test of the SEK-onto-USD POST is **not** done: the browser extension disconnected mid-verification. The pre-fix reproduction was confirmed on a local stack (USD invoice ended at rate 11.24); the post-fix confirmation is outstanding and I'll add it here once I can drive the app again.

### DIFF
--- a/apps/erp/app/routes/x+/purchase-invoice+/$invoiceId.exchange-rate.tsx
+++ b/apps/erp/app/routes/x+/purchase-invoice+/$invoiceId.exchange-rate.tsx
@@ -3,7 +3,7 @@ import { requirePermissions } from "@carbon/auth/auth.server";
 import { flash } from "@carbon/auth/session.server";
 import type { ActionFunctionArgs } from "react-router";
 import { redirect } from "react-router";
-import { getCurrencyByCode } from "~/modules/accounting";
+import { resolveCurrencyAndRate } from "~/modules/accounting";
 import {
   getPurchaseInvoice,
   isPurchaseInvoiceLocked,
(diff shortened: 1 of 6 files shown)


BettrCallRaul:
Runtime verification is now done — the outstanding item in the PR description is closed.

On a local stack with a EUR-base company, posting `currencyCode=SEK` to a sales invoice denominated in USD:

- **Before the fix:** the invoice ended at `currencyCode` USD with `exchangeRate` 11.24 — SEK's rate stamped onto a USD document, mispricing every line through the `converted*` columns.
- **After the fix:** the invoice stays at USD @ 1.085, header and line, with `convertedUnitPrice` 1,953,000 (1,800,000 × 1.085). The posted currency is ignored.

The legitimate path still works: the refresh button in the properties panel behaves as before, since the server now reads the currency the UI was already sending.

PR #1543, opened 2026-09-01
fix(purchasing): convert supplier part prices to the supplier's currency, not away from it

BettrCallRaul:
## Problem

`supplierPart.unitPrice` and `supplierPartPrice.unitPrice` are stored in the company **base** currency. Neither table has a currency column, and all three writers put base there — `convert/index.ts:1767` (`supplierUnitPrice / exchangeRate`), supplier-quote `$id.finalize.tsx:187`, and `SupplierPartForm.tsx:188`, which formats the field with `baseCurrencyCode`.

The field they prefill, `supplierUnitPrice`, is in the **supplier's** currency (`PurchaseOrderLineForm.tsx:716` formats it with `orderCurrency`). The document's `exchangeRate` is foreign-per-base, so base → supplier is **multiply**.

Both purchasing line forms divided, and `resolveSupplierPrice` divided again on the price-break path. The value written into a supplier-currency column was `base / rate` where it should be `base × rate` — wrong by the rate squared.

Nobody notices a plausible small number in a foreign-currency field. The generated base column then derives from it, so receiving the line creates a cost layer at the wrong magnitude and drags the weighted-average item cost with it.

## Fix

`resolveSupplierPrice` now takes a base fallback and base breaks and returns the supplier-currency price. Both call sites hand it the raw base value.

Also: the `itemData` initializer seeded `fallbackUnitPrice` from `initialValues.supplierUnitPrice` — a supplier-currency value — into a field the helper treats as base. It now divides to base, so the field means one thing throughout. That inconsistency predates this PR but the fix makes the direction explicit, so leaving it would have been a new bug.

## Verification

On a local stack, base EUR, supplier Deep Space RF in SEK at rate 11.24, part RW-010 with a stored base price of €14,500:

Adding that item to a SEK purchase order now defaults **Unit Price: SEK 162,980.00** (14,500 × 11.24). Before the fix it defaulted to SEK 1,290.04 — understating the line by a factor of 126.

Typecheck and lint clean. Found by an audit of the costing and supplier-pricing surfaces.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected supplier price calculations on purchase orders and purchase invoices when exchange rates are applied.
  - Improved price-break handling so prices are resolved consistently in the base currency before conversion to the supplier’s currency.
  - Prevented incorrect unit prices caused by duplicate or misplaced currency conversions.
  - Improved the accuracy of fallback pricing for supplier parts and item costs.

### DIFF
--- a/apps/erp/app/modules/invoicing/ui/PurchaseInvoice/PurchaseInvoiceLineForm.tsx
+++ b/apps/erp/app/modules/invoicing/ui/PurchaseInvoice/PurchaseInvoiceLineForm.tsx
@@ -149,7 +149,11 @@ const PurchaseInvoiceLineForm = ({
     taxAmount: initialValues.supplierTaxAmount ?? 0,
     taxPercent: initialValues.taxPercent ?? 0,
     priceBreaks: [],
-    fallbackUnitPrice: initialValues.supplierUnitPrice ?? 0
+    // fallbackUnitPrice is BASE currency, matching what resolveSupplierPrice
+    // expects. initialValues.supplierUnitPrice is the SUPPLIER's, so divide.
+    fallbackUnitPrice:
+      (initialValues.supplierUnitPrice ?? 0) /
+      (routeData?.purchaseInvoice?.exchangeRate || 1)
   });
 
   // Re-derive the tax amount when the line's base changes — never on mount, so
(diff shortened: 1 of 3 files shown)


BettrCallRaul:

Why this matters today

- Conditions: a supplier whose currency differs from base, and an item with a supplier part price or price breaks. Those prices are stored in base currency.
- Steps: create a purchase order for that supplier and add the item. The unit price prefills as base divided by the rate instead of multiplied, so it is wrong by the rate squared. Price breaks divide a second time. Editing an existing line seeds the fallback price from the supplier-currency value as if it were base, so the error repeats on edit. The purchase invoice line form has the same two paths.
- Impact: a plausible small number sits in a foreign-currency field and nobody questions it. If saved, the line's generated base price is wrong, the receipt creates a cost layer at that value, and the item's weighted average cost drifts.
- Example: EUR base, SEK supplier at 11.24, part stored at 14,500. Prefill is SEK 1,290 instead of SEK 162,980.

PR #1548, opened 2026-09-01
fix(documents): email the line unit price in the currency the email is formatted in

BettrCallRaul:
## Problem

Both order emails build their money formatter from the **document's** currency — each carries a comment saying exactly why (`PurchaseOrderEmail.tsx:67`, `SalesOrderEmail.tsx:66`: *"a supplier priced in JPY must not be emailed as `$20.00`"*) — and then format a **base-currency** column with it.

- `SalesOrderEmail.tsx:269` printed `salesOrderLine.unitPrice`, stored in BASE, while the line total beside it comes from `getLineTotal` → `getLineSubtotal`, which sums `convertedUnitPrice` (`utils/sales-order.ts:47`). The customer saw a base unit price labelled with their currency next to a correctly converted total — quantity × unit price did not equal the total on the same row.
- `PurchaseOrderEmail.tsx:272` printed `purchaseOrderLine.unitPrice`, the **generated base** column, under the supplier's currency, while its `getLineTotal` sums `supplierUnitPrice` (`utils/purchase-order.ts:49`). The PO **PDF** has always used `supplierUnitPrice` for this column (`pdf/blocks/purchaseOrder/LineItemsBlock.tsx:163`).

`SalesInvoiceEmail.tsx:256` already used `convertedUnitPrice`. Its two siblings were missed — the formatter's currency was corrected in all three, the amount column in only one.

These go to customers and suppliers, so the wrong number left the building.

## Fix

Each email now formats the same column its own total helper sums: `convertedUnitPrice` for the sales order, `supplierUnitPrice` for the purchase order.

## Verification

`tsc` on `packages/documents`: 104 pre-existing errors before and after, none in either changed file. There are no tests over the email templates, and no snapshot fixtures to extend.

Not exercised end to end — sending an order email from the test instance would email a real address. The claim rests on the column families, which are established by each file's own total helper doing it the other way.

Found by an audit of the document layer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Purchase order emails now display supplier-quoted unit prices, including valid zero-priced items.
  * Sales order emails now show unit prices in the document’s currency.
  * Sales invoice emails now show document-currency unit prices.
  * Sales order comment lines continue to display “-”.
  * Low-value unit prices are now displayed accurately without rounding to zero.
  * Settlement subtotals and totals continue to use standard money formatting.

### DIFF
--- a/packages/documents/src/email/PurchaseOrderEmail.tsx
+++ b/packages/documents/src/email/PurchaseOrderEmail.tsx
@@ -268,8 +268,12 @@ const PurchaseOrderEmail = ({
                   <Text className="text-xs font-semibold">
                     {line.purchaseOrderLineType === "Comment"
                       ? "-"
-                      : line.unitPrice
-                        ? formatter.format(line.unitPrice)
+                      : // `formatter` is the SUPPLIER's currency (above), and
+                        // purchaseOrderLine.unitPrice is the GENERATED BASE
+                        // column -- supplierUnitPrice is what the supplier
+                        // quoted, and it is what getLineTotal already sums.
+                        line.supplierUnitPrice
+                        ? formatter.format(line.supplierUnitPrice)
                         : "-"}
                   </Text>
                 </Column>
--- a/packages/documents/src/email/PurchaseOrderEmail.tsx
+++ b/packages/documents/src/email/PurchaseOrderEmail.tsx
@@ -280,7 +280,10 @@ const PurchaseOrderEmail = ({
                         // purchaseOrderLine.unitPrice is the GENERATED BASE
                         // column -- supplierUnitPrice is what the supplier
                         // quoted, and it is what getLineTotal already sums.
-                        line.supplierUnitPrice
+                        // A zero unit price is a real price (a no-charge
+                        // line), not a missing one, and getLineTotal already
+                        // sums it as 0 -- so test for absence, not falsiness.
+                        line.supplierUnitPrice != null
                         ? rateFormatter.format(line.supplierUnitPrice)
                         : "-"}
                   </Text>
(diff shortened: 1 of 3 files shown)


BettrCallRaul:
Second commit added. An adversarial verification pass over the original finding turned up a **second defect on the same line**, which the first commit did not address.

All three order emails formatted the unit-price column with `getMoneyFormatter`, whose maximum equals the currency's decimals. Per `.claude/rules/numeric-precision.md` a per-unit price is the **rate** kind — the currency's decimals are its floor and `SCALE` its ceiling — precisely so a sub-cent price does not print as `0.00`. That is the documented [#1203](https://github.com/crbnos/carbon/issues/1203) class.

The PDFs already split the two kinds: `getRateFormatter` for the unit-price column, `getMoneyFormatter` for totals, tax and shipping (`pdf/blocks/purchaseOrder/LineItemsBlock.tsx:163` vs its Summary block). The emails used money for everything.

`SalesInvoiceEmail` is now in scope too. I said in the description that it "already used `convertedUnitPrice`" and was correct — that was right about the *currency* and wrong about the *formatter kind*. A 0.164 unit price emailed as 0.16 from all three.

Totals, tax and shipping stay on the money formatter; those are settlement amounts, not rates.

`tsc` on `packages/documents`: 104 pre-existing errors before and after, none in the three changed files.

BettrCallRaul:

Why this matters today

- Conditions: a sales order or purchase order in a currency other than base, or any document with a per-unit price below the currency's smallest unit.
- Steps: send the order or invoice email from the document. On a foreign-currency sales order the unit price column is the base number labelled with the customer's currency, while the line total beside it is converted, so quantity × unit price does not equal the total on the same row. The purchase order email does the same with the supplier's currency. All three emails round unit prices to the currency's decimals, so 0.164 prints as 0.16 while the PDF prints 0.164. A zero-price purchase order line prints a dash.
- Impact: these leave the building. Customers and suppliers receive an email whose rows do not add up, or whose unit price differs from the PDF attached to it.

PR #1555, opened 2026-09-02
Feat/docs link visibility

aashu0148:
## What does this PR do?

- The API docs showed `rest.carbon.ms` as the host in every endpoint URL, code sample and MCP
  snippet. That address only works for customers on Carbon Cloud, so anyone self-hosting or on a
  non-default region was being handed copy-paste commands that point at a server they cannot
  reach, with nothing on the page saying so.
- Readers whose instance we cannot identify (most people, since they arrive from search) now see
  `<your-host>` in place of the address, as a clickable chip that opens the existing API
  configuration dialog. Once they set their host, every URL, code sample, MCP endpoint and
  Settings link on the page uses it. The paths themselves stay real in all cases.
- Opening the API docs from inside the app now carries that instance's own API address across, so
  signed-in users skip the configuration step entirely.
  
  
  ## Video
  

https://github.com/user-attachments/assets/52cf5edc-4920-4ed6-b4c0-2bd1f06f0202

  

## Summary by CodeRabbit

* **New Features**
  * Added host visibility across API documentation, including a clickable `<your-host>` placeholder for unconfigured instances.
  * Added URL-based host configuration and a “Not sure” mode in the configurator.
  * Documentation links, code samples, and MCP endpoints now reflect configured hosts and preserve path prefixes.
* **Bug Fixes**
  * Safely escaped displayed host values.
  * Improved configurator behavior with multiple controls.
  * Prevented credentials from being exposed through link-provided or non-secure hosts.

### DIFF
--- a/.ai/plans/2026-09-02-docs-host-visibility.md
+++ b/.ai/plans/2026-09-02-docs-host-visibility.md
@@ -1,0 +1,569 @@
+# Docs Host Visibility — implementation plan
+
+**Spec / source:** `.ai/specs/2026-09-02-docs-host-visibility.md`
+**Branch:** `feat/docs-link-visibility`
+**Worktree:** `/Users/aashu/work/carbon/carbon-feat-docs-link-visibility`
+**Mode:** ship-it autonomous
+
+## Progress
+- [x] Task 1: Extract `parseBaseUrl` into a shared module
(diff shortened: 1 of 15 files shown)


sidwebworks:
We have an EU instance and an ITAR instance - I wonder if we should make it aware of those? @barbinbrad

sidwebworks (approved):
LGTM

PR #1562, opened 2026-09-03
refactor(currency): platform-global exchange rates with per-company overrides

barbinbrad:
## What does this PR do?

Replaces the group-scoped `currency.exchangeRate` column and the integration-gated nightly job with a platform-global, effective-dated `exchangeRate` store (USD-anchored, seeded with a bootstrap snapshot, written daily for everyone — the `exchange-rates-v1` integration is deleted), per-company `exchangeRateOverride` pins that always beat the feed, and one SQL resolver `get_exchange_rate(companyId, code)` (base → 1, override, else the ratio of two market rates; a missing rate is an error, never 1). `currency` survives as group-scoped config (decimalPlaces/active/historicalExchangeRate); document snapshot semantics are unchanged, all ~40 stamping sites resolve through the resolver, refresh routes take the currency from the document rather than the request, the missing salesInvoice header→line propagation trigger lands, rate > 0 CHECKs land on all 12 document tables (with in-migration repair of legacy NULL/zero rates), and consolidation closing/average rates come from the global store as explicit target-per-source pair ratios. Design, blast-radius research, and all resolved decisions: `.ai/specs/2026-09-02-currency-exchange-rate-refactor.md`.

Relationship to the open exchange-rate fix PRs (full audit in `.ai/research/2026-09-02-exchange-rate-pr-audit.md`): **supersedes** #1542; **subsumes** #1526, #1527, #1530, #1541 (guarded migrations, safe if those land too); **independent — can merge either order**: #1532, #1543, #1544, #1547, #1548, #1549.

Note for reviewers: `packages/jobs/manifests/schema.json` is regenerated from a clean fully-migrated schema — the copy on main contains `onshapeSyncRun`/`onshapeItemSyncState`, tables no migration defines (the baseline was committed from a local DB carrying unmerged work), so the backup pre-commit check was skipped for these commits with that documented cause.

No screenshots yet — the browser pass (rates page provenance badges, override/reset, foreign-currency quote stamping, invoice line cascade) is queued; will attach before merge.

- [x] I have self-reviewed the code (strict multi-agent self-review; all must-fix findings applied in the second commit).
- [x] I confirm automated tests are in place that prove my fix is effective or that my feature works (new `update-exchange-rates.test.ts` pins the USD-anchoring math; backup scope tests, `@carbon/checks` conformance gate, and the full suite pass).

## How should this be tested?

- Run `pnpm db:migrate` (four new migrations), then `pnpm run generate:types`.
- No env setup needed: rates work with no integration and no API key (seeded snapshot); with `EXCHANGE_RATES_API_KEY` set, the nightly `update-exchange-rates` job refreshes the global store (one API call per installation).
- Happy path: on a USD-base company, `SELECT get_exchange_rate('<companyId>', 'EUR')` returns ~0.92; the company's own base returns exactly 1; an unknown code errors with a remediation message. In the app: Accounting → Exchange Rates shows resolved rates with Market/Your-rate/Base badges; editing a rate creates an override (survives the feed), Reset to market rate deletes it; creating a quote in a foreign currency stamps a real rate (never 1); a sales-invoice rate refresh cascades to its lines; Settings → Integrations no longer lists Exchange Rates.
- Two companies in one group with different base currencies resolve the same foreign code to different, correct rates — the bug class behind #1525/#1542.

## Checklist

- I have read the [contributing guide](https://github.com/crbnos/carbon/blob/main/.github/CONTRIBUTING.md)

## Summary by CodeRabbit

- **New Features**
  - Added global market exchange rates with company-specific overrides.
  - Users can view rate sources and update, save, or reset currency overrides.
  - Added opening-balance posting workflows.
  - Exchange-rate setup now links directly to the exchange-rates screen.

- **Bug Fixes**
  - Missing or invalid rates now produce clear errors instead of silent defaults.
  - Improved exchange-rate propagation across invoices and order lines.
  - Consolidated reports now show an error when currency translation fails.
  - Documents consistently use their recorded exchange rates.
  - Referrer handling now helps prevent unsafe redirects.

- **Localization**
  - Added currency and exchange-rate translations across supported languages.


barbinbrad:
Addressed the CodeRabbit review in 1afafcc. Disposition per finding:

**Fixed in code**
- **NaN passes `NUMERIC > 0`** — added `AND <> 'NaN'::numeric` to the rate CHECKs on `exchangeRate`, `exchangeRateOverride`, and all 12 document tables (verified: an authenticated NaN insert now fails the constraint).
- **Provenance in the override migration** — the predicate now excludes any group that *ever* had an `exchange-rates-v1` row (active or not); only never-fed groups' hand-edited rates become pins.
- **`exchangeRateHistory` drop** — now guarded: refuses with an explicit error if any deployment holds rows, instead of discarding.
- **Orphaned `supplierInteraction`** — both PO create paths resolve the rate *before* inserting the interaction.
- **`upsertExchangeRateOverride`** — 23505 retry closes the concurrent first-pin race; signature now takes `createdBy`/`updatedBy` so the MCP dispatcher's `injectAuth` reaches it (tool metadata regenerated).
- **Open redirect (CWE-601)** — fixed centrally: `requestReferrer` reduces the Referer to a same-origin relative path or null, covering all ~57 call sites, not just this route.
- **`error: null` success on null rate** — the three route branches now match the service idiom (the RPC's generated type is non-nullable `number`, so the null arm was unreachable dead code).
- **Base-change atomicity** — company update + override cleanup now run in one Kysely transaction (`updateCompanyWithBaseCurrencyChange`).
- **`create` USD default** — a supplier with no currency now resolves to the company's base currency (rate 1 by definition), across the whole function.
- **get-method parity fallback** — a legacy null line rate now falls back to the *source quote's stamped header rate* before 1.
- **NULL re-entry via the salesInvoiceLine trigger** — COALESCE-guarded (header is NOT NULL, but the trigger can no longer write NULL regardless).

**Addressed with rationale (documented in-file/spec rather than changed)**
- **Same-day upsert vs "insert-only"** — the spec now states the real contract: the feed may correct the same day's row; user data never lives in that table, so nothing is clobbered.
- **Par repair on foreign-currency rows** — kept, with the reasoning in the migration: pre-refactor code read these NULL/zero rates as 1 everywhere, their line snapshots defaulted to 1, and no truer rate is recoverable; re-deriving from today's market would rewrite historical documents.
- **CHECK not rejecting NULL** — deliberate: nullable header columns accept rate-less inserts from pre-stamping writers; NULL is handled at the resolver/stamping layer. The one trigger path back to NULL is now guarded.
- **Resolver SECURITY INVOKER** — deliberate; the authorization contract (employees of the company or service role) is now documented on the function. DEFINER would trade RLS for a hand-rolled tenant check.
- **`historicalExchangeRate` anchor** — documented as target-per-source against the group presentation currency by convention (all consolidated reports pass the root base); re-anchoring a manually stated IAS-21 rate is a consolidation follow-up, not a silent conversion.

PR #1570, opened 2026-09-04
feat(settings): regenerate an API key's secret in place

BettrCallRaul:
## Why

A key is shown once at creation and Carbon stores only its hash. Lose it or leak it and the only path was delete-and-recreate: re-enter the name, re-grant every scope, rewire every integration to a new key id. The docs literally said so.

## What

**Settings → API Keys → row menu → Regenerate API Key** issues a new secret on the existing row. Name, scopes, expiry and id are untouched — only `keyHash`/`keyPreview` change — so integrations keep their permissions and lose their credential. The old value stops authenticating on the next request: every resolver (`auth.server.ts`, the three in `functions/lib/supabase.ts`, the two SQL functions) hashes the incoming header per request and nothing caches the lookup.

- Confirm dialog with a destructive alert and the key's last-used age; reveal dialog (extracted to `ApiKeyView`, shared with creation) with a "previous key has stopped working" notice.
- `regenerateApiKey` refuses an expired key — a fresh secret on an expired row is rejected on its first request — and the modal explains and disables the button. Extending `expiresAt` via Edit is untouched and still revives the same secret; the two operations stay independent.
- New `apiKey.updatedAt` (migration `20260904073608`); regenerate and edit stamp `updatedBy`/`updatedAt`. `lastUsedAt` is cleared on regenerate since it described the retired secret.
- `mintApiKey()` (`settings.server.ts`) is now the single place a secret is generated, used by creation and regeneration.
- `settings_regenerateApiKey` is added to `MCP_BLOCKED_TOOL_NAMES`: it takes the hash from the caller, which is fine for the route but would let a company-scoped token overwrite any key's secret over MCP.
- `ApiKeysTable` no longer renders its own `<Outlet/>` — the route already does, and the pair mounted every child modal twice.
- Docs, agent KB, module `AGENTS.md`, Lingui catalogs (English only; other locales left for `pnpm translate`).

## Verified

Browser-driven on a local stack, with curl against PostgREST (`carbon-key`) and `/api/mcp` (Bearer) before and after each step:

- create → old key 200 on both paths → regenerate → old key rejected on both, new key 200 on both; row shows new preview, `updatedBy`/`updatedAt`, `lastUsedAt` null, scopes intact
- key expired via SQL → red badge; modal shows the extra alert and disables; POSTing the action directly leaves `keyHash` unchanged
- Edit → expiry extended → same hash, key live again → regenerate → new secret works, extended expiry survives

`turbo typecheck --filter=erp`, biome, `@carbon/checks` (87/87 incl. baseline), `db:check:datasets`, `db:check:backups` all green.

## Noticed, deliberately not in this PR

- `WebhooksTable` has the same duplicate `<Outlet/>` as `ApiKeysTable` had.
- `lastUsedAt` never stamped during testing, even via `/api/mcp` which calls `requirePermissions` — the fire-and-forget at `auth.server.ts:269` deserves a look; REST calls bypass app code entirely so never stamp it.
- `get_companies_with_permission` (`20250127164617`) still reads `apiKey."key"` and the `api-key` header; both were dropped/renamed in `20260219162954`.
- Docs say expiry "can't be extended on a used key"; the edit form does exactly that.
- `tool-metadata.json` shrinks by 177 lines: the committed `settings_upsertApiKey` schema carried 50 company-settings params that main's own generator no longer emits.

https://claude.ai/code/session_01JJU5TyZDYmw5mVvHqs8Bqq

## Summary by CodeRabbit

* **New Features**
  * Added the ability to regenerate API keys without changing their name, scopes, expiration, or identifier.
  * Regeneration immediately invalidates the previous secret and displays the new secret only once, with a copyable MCP command.
  * Expired keys cannot be regenerated; key usage and regeneration details are shown during confirmation.

* **Documentation**
  * Added guidance for regenerating keys and updating integrations after rotation.
  * Updated troubleshooting instructions to recommend regeneration instead of deleting and recreating keys.


barbinbrad:
Why is this better than creating a new one on leak?

BettrCallRaul:
@barbinbrad it not really for security its for convenience if you lose the key that you've generated. Convenience in the fact that you don't need to delete and recreate you can just regenerate and sunset the old one.

PR #1583, opened 2026-09-06
feat(learn): Carbon Learn — role tracks, verified hands-on challenges, and certification

sidgaikwad:
Carbon teaching Carbon. A learner reads a documentation page, answers scenario
questions about it, then proves the skill by doing the real thing in their own
company — and Carbon goes and checks. Passing a track's exam issues a
certificate with a public verification link.

Spec: `.ai/specs/2026-08-29-learning-paths-certification.md`
Plan: `.ai/plans/2026-09-06-learning-paths-certification.md`

## What ships

**Nine tracks** — Fundamentals plus one per role — with **780 questions** and
**24 hands-on challenges** verified against the learner's own records. Each
track lives in its own file under `learn/tracks/`, so a content change reviews
on its own rather than as a diff buried in a 2,000-line curriculum.

**Grading is server-side and version-bound.** The answer key never reaches the
browser. Every answer is graded at submission against the bank for the attempt's
own `LEARN_CONTENT_VERSION`; if the curriculum has moved on mid-attempt the
attempt is voided rather than graded against questions the learner never saw.

**A challenge is not a checkbox.** `Check my work` re-reads the database under
the session's own company, scoped to records the learner created after the
*server* wrote their start time, and reports the first unmet requirement by
name — "PO000028 is still Draft — release it from the order's status menu".
Retries are unlimited and never penalised; the fix-and-recheck loop is the
lesson. Checkers are read-only and never create, edit, or delete anything.

**The motivation layer is the learner's alone.** XP is an append-only ledger
with a uniqueness constraint, so a retry can never inflate a total. Streaks are
weekly, not daily — a weekend cannot break one. There are no leaderboards.
Managers see a projection: assignment status and certificates, never XP,
streaks, activity, or which questions anyone missed.

**Engine tables carry no write policies at all** (`learnAttemptAnswer` has RLS
enabled with zero policies — service-role only), following the
`itarCertification` precedent. The engine is the only writer.

## Also here

- A **question-performance report** at `/x/resources/learn-questions`. A
  question people keep failing is usually a documentation problem, so it sorts
  worst-first and links the page to go and fix. It aggregates only, and hides
  any question with fewer than five attempts — below that, a correct-rate on a
  small team identifies one named person.
- An **anonymous `<Quiz>` component** for the docs site — localStorage only, no
  backend — seeded into three pages. `generate-agent-kb.ts` strips quiz blocks
  entirely, so the in-app agent is never taught the deliberately-plausible wrong
  answers.
- All twelve non-source locales translated.

## Three schema assumptions in the plan were wrong

Every checker was written against the live schema rather than the plan's
assumed column names, and three of those assumptions did not survive contact:

| Assumed | Actually |
|---|---|
| `salesOrder.quoteId` | No such column — conversion preserves `opportunityId` |
| `accountingPeriod.status` is the close state | That column is Active/Inactive; the close state is `closeStatus` (`Open \| Locked \| Closed`) |
| `employeeType.createdBy` | Does not exist; `employeeTypePermission` has no `companyId` either |

All three are recorded in `.claude/rules/learn-system.md` alongside the
already-documented `purchaseOrderLine.purchaseOrderId` trap, because each is the
same shape of bug: matching on the obvious name silently matches nothing.

## Verification

```
230 learn tests           (103 of them proving the checkers can actually fail)
 87 conformance checks
  4 dataset checks
    typecheck + lint clean across erp, database, jobs, docs
    linguito check: 0 missing in all 12 locales
```

Every checker ships with tests proving it **fails on an empty company** naming
its first requirement, **passes on a known-good sequence**, and **honours its
scope** — a checker that cannot fail certifies nothing.

Verified in-browser end to end: all nine tracks live on the hub, a quiz on a new
track drawn from its pool, graded server-side, and returned with per-question
explanations and documentation links; a retake drawing a different form; the
question report correctly empty behind its five-attempt privacy floor.

## Reviewing this

The content is large but mechanical — the nine `tracks/*.ts` and nine
`banks/*.server.ts` files are data. The load-bearing code is
`learn/engine.server.ts` (the only writer), `learn/checkers/reader.server.ts`
(the only database surface a checker may touch), and the migration's RLS.

## Summary by CodeRabbit

- **New Features**
  - Added Carbon Learn with role-based learning tracks, quizzes, hands-on challenges, XP, badges, streaks, certification exams, renewals, and downloadable certificates.
  - Added learner dashboards, progress tracking, activity heatmaps, weekly goals, certificate verification, and administrator assignment/team management tools.
  - Added learning notifications, certificate expiry alerts, question-performance reporting, and multilingual interface support.
- **Bug Fixes**
  - Improved exam answer reset behavior and challenge validation accuracy.
  - Prevented invalid or outdated learning attempts from being graded.
- **Documentation**
  - Added learning-system guidance and research/specification documents.


sidgaikwad:
Worked through all 17 findings. Both Critical ones were real. **12 fixed**, **4 declined**, 1 partially — reasons below.

Merged `main` first (Zod 3.25 → 4.5 landed; the three Learn validators using `errorMap` were the last call sites of the old form in the repo) and regenerated the conflicting `tool-metadata.json` — 1,513 tools, all 1,495 from main intact.

### Fixed

| # | Finding | Note |
|---|---|---|
| 8, 7, 10 | Three missing `toIso` sites | Confirmed all three. `expiresAt` compared as a `Date` against an ISO string meant the 45-minute exam limit **never fired**; `submittedAt` sorted as `Date` objects orders by weekday *name*, so a cooldown could anchor to an older attempt |
| 13 | Exam `selected` not reset per question | Confirmed. The `key` on the `<form>` doesn't reset parent state |
| 9 | `finalizeExamAttempt` graded voided attempts | Confirmed — contradicted the documented invariant |
| 14 | Verification link never rendered | Confirmed. `issueCertificate` now returns `{ id, verificationCode }` |
| 12 | Five-attempt floor bypassed by the MCP tool | Confirmed and moved into the service; constant now in `gamify.ts` with the other numbers |
| 11 | `zfd.repeatable` for `groupIds` | Confirmed against the repo's own convention |
| 15 | Cross-company group via PostgREST | Added `learnAssignment_groups_in_company`. Verified against the local DB in a rolled-back transaction: refuses on INSERT *and* UPDATE, accepts same-company groups |
| 5 | Permissions counted for any company | Now counts only grants naming this company or the `"0"` wildcard |
| 4 | Unset reordering policy read as a planning policy | Now an allowlist, with a regression test. Worth noting the column is `NOT NULL`, so this was not reachable from real data — defence-in-depth rather than a live bug |
| 17 | Notification lookups unscoped | Both scoped by company; fail closed when it is absent |
| 16 | PDF styling conventions | Now `useTw()` + `getSafeFontFamily()` + semantic grays |

### Partially fixed

**6 — challenge attribution.** The supplier half is fixed: the fallback now reaches only rows with *no* recorded author, never one demonstrably created by a different user, so the capstone can't pass on a colleague's work. The `employeeType` half I've left — that table has no `createdBy` column at all, so "attributable to the learner" isn't expressible without a schema change, and adding an audit column to a shared table for a training checker is a bigger call than this PR should make. The limitation is documented in the rule file.

### Declined

**1, 2, 3** are findings against `.ai/specs/…`, a design document that was reviewed and merged separately in #1509 — not code in this PR.

On **2** specifically, I'd push back on the substance too: `learnXpEvent`'s policy is `userId = auth.uid() AND companyId = ANY(get_companies_with_employee_role())`, which is the standard policy shape used across this repo. The data it exposes is the learner's *own* XP in a company they are already a member of and can switch into — not another tenant's. Narrowing just these tables would deviate from the house convention without closing a real gap.

### Verification

231 learn tests (up 1), 100 conformance checks, 4 datasets, typecheck clean across `erp`, `@carbon/documents`, `@carbon/jobs`, `@carbon/database`, and all 12 locales still at zero missing.

huanglinfei091-cmd:
Hi @sidgaikwad — I noticed the `Catalog` job failed at `Run pnpm run check:workflow-catalog` in [this GitHub Actions run](https://github.com/crbnos/carbon/actions/runs/34079401006) for this PR.

I’m testing [RunBack](https://github.com/huanglinfei091-cmd/runback), an early alpha CLI that reconstructs a failed public Actions run locally with `act`. For this run, the useful question is whether the workflow-catalog check reports the same mismatch, or whether reconstruction stops with an honest blocked result.

```bash
runback doctor
runback https://github.com/crbnos/carbon/actions/runs/34079401006
```

The Linux amd64 binary and checksum are in the [v0.1.0-alpha release](https://github.com/huanglinfei091-cmd/runback/releases/tag/v0.1.0-alpha). If you try it, the verdict (`SAME_FAILURE`, `DIFFERENT_FAILURE`, or `REPLAY_BLOCKED`), Stage/Cause, and any install friction would be useful. Please do not post tokens or secrets. I’m asking for candid alpha evidence rather than a star or promotion; a failed reproduction is useful too.

PR #1595, opened 2026-09-08
fix(ui): truncate long titles in BoM/BoP rows 

aashu0148:
## What does this PR do?

- Fixes long operation/material titles in the Bill of Process and Bill of Materials cards pushing the time badges off the edge of the card, where they were clipped and overlapped the settings icon.
- Long titles now truncate with an ellipsis so the badges and row actions always stay fully visible. Applied to all six places that use these cards: job, quote, and item methods (both BoM and BoP).

## Summary by CodeRabbit

- **Style**
  - Improved text truncation and responsive layout for item names and operation descriptions across bills of materials, bills of process, production jobs, and quotes.
  - Long titles now fit more reliably within list rows without forcing surrounding content beyond available space.
  - Preserved checked-state styling and existing pointer interactions.

### DIFF
--- a/apps/erp/app/components/SortableList.tsx
+++ b/apps/erp/app/components/SortableList.tsx
@@ -144,7 +144,7 @@ function SortableListItem<T>({
                         {typeof item.title === "string" ? (
                           <span
                             className={cn(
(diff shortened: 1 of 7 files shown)


naveenkash (approved):
Lgtm

PR #1622, opened 2026-09-11
Harden the generator and utility scripts

parthingle:
- generate-db-types: refuse non-local database URLs without echoing them,
  generate into temporary files and validate (parses, exports Database) before
  atomically replacing both output copies, restore the last-good files on
  failure, and sort each table's Relationships block so the output is a pure
  function of the schema (supabase gen types emits catalog order, which differs
  between databases built from the same migrations).
- generate-swagger-docs: verify the partners view definition through Studio's
  catalog query and normalise only PostgREST's known duplicate-alias primary-key
  annotation, so two independently built databases produce identical output.
- generate-mcp: drop the wall-clock timestamp from the metadata (repeatable
  output) and disable turbo caching for //#generate:mcp until its transitive
  inputs are declared.
- sandbox / model-upload / sales-invoice-report: read configuration from
  explicit local environment variables instead of embedded values; error
  messages name the missing variable without printing values.
- Tests for all of the above under scripts/lib (node:test, run with tsx).

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

## Summary by CodeRabbit

- **New Features**
  - Added environment-based configuration for local tools, model uploads, sandbox queries, and sales invoice reports.
  - Added validation and safer handling for database type and Swagger documentation generation.
  - Added support for configurable report output paths and home-directory paths.

- **Bug Fixes**
  - Improved error reporting and exit status handling for script failures.
  - Preserved existing generated files when generation or validation fails.
  - Prevented sensitive configuration values from appearing in error messages.

- **Tests**
  - Added comprehensive coverage for configuration validation, generation failures, schema verification, and output preservation.

### DIFF
--- a/scripts/generate-db-types.ts
+++ b/scripts/generate-db-types.ts
@@ -1,102 +1,22 @@
-import {
-  closeSync,
-  copyFileSync,
-  openSync,
-  readFileSync,
-  writeFileSync
-} from "node:fs";
-import { spawnSync } from "node:child_process";
-import { join } from "node:path";
-import * as dotenv from "dotenv";
+import { existsSync, readFileSync } from "node:fs";
+import { parseEnv } from "node:util";
+import { generateDatabaseTypes } from "./lib/generate-db-types";
 
-dotenv.config({ path: ".env" });
-dotenv.config({ path: ".env.local", override: true });
-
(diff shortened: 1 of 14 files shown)


sidwebworks:
Thanks for the PR 🙌🏼 
I'll make some changes to it in a follow up PR

PR #1660, opened 2026-09-17
(feat): see past already-installed parts in the 3D assembly viewer

aashu0148:
## What does this PR do?

- Adds a visibility control to the 3D assembly viewer so an operator can see past the parts that are already fitted. Deep into a build the finished assembly buries the parts the current step is asking for, and until now there was no way to look through it.
- The control offers four named views instead of raw toggles: **Build** (what is on the bench right now), **Focus** (already-fitted parts fade to see-through), **Isolate** (only the current step's parts), and **Full** (the finished product). Isolate answers the question the viewer exists for: "just show me the part I am fitting."
- The camera now stops dodging parts it is not drawing. Previously it could hide a part visually while still angling the view around it, so hiding the clutter did not actually give you a clear look at the step.
- Same control in both the ERP assembly editor and the MES shop floor screen, since an author and an operator ask the same question of the model.
- Existing behavior is unchanged by default: Build is the default view and matches what the viewer did before.
- Documents the guided assembly view and the four views in the MES reference page, which had no coverage of assembly playback at all before this.

## image 
image

## How should this be tested?

- **Environment variables:** None.
- **Minimal test data:** A published assembly instruction with a 3D model and several steps, ideally one where later steps are buried behind earlier ones (that is where the change matters).
- **Expected (happy path):** Open the assembly in the ERP editor (`/x/assembly/<id>`) or on an MES job operation, step partway into the build, and use the view buttons in the player toolbar. Build shows the assembly as built so far, Focus fades the already-fitted parts to see-through, Isolate leaves only the current step's parts, and Full shows everything solid. On Isolate the camera should frame the current step's parts directly rather than angling around parts that are no longer drawn.
- **Other info:** Unit tests cover the rendering rules and the camera scoring, including the invariant that anything drawn as hidden is never treated as an obstacle by the camera. Run with `pnpm --filter @carbon/viewer test` (31 tests in `visibility.test.ts`; 120 across the package).

## Summary by CodeRabbit

* **New Features**
  * Added four assembly viewer modes: **Build**, **Focus**, **Isolate**, and **Full**.
  * Installed components can now appear solid, ghosted, or hidden.
  * Camera framing now avoids hidden components and reduces the impact of ghosted components.
  * The view controls remain usable in component-picker mode and wrap on narrow panels.
* **Documentation**
  * Added MES and ERP guidance describing the new assembly views and their effect on visualization.
  * Clarified that view changes do not alter assembly steps, installed components, or reported data.


.ai/specs/2026-09-17-viewer-installed-component-visibility.md
```diff
@@ -0,0 +1,432 @@
+# Installed-component visibility in the assembly viewer
+
+> Status: draft
```
naveenkash: Is it done?

naveenkash (approved):

PR #1663, opened 2026-09-17
perf(erp): stop the app shell downloading whole tables from the browser

sidwebworks:
## What does this PR do?

The authenticated ERP shell mounts `RealtimeDataProvider`, which pulls six whole tables straight from the browser into nanostores on every page load. This PR fixes three real costs in that path — it does **not** stop the shell downloading whole tables, which needs server-side search comboboxes and is deliberately out of scope here (see below).

1. **`fetchAllFromTable` requested an exact count on every page.** PostgREST ran a `COUNT(*) OVER ()` across the full filtered set per page, while only page 0's count was kept and no caller anywhere read it. It also paged serially, so a 150k-row table was 150 sequential round trips. Now: no count, a query factory, first page alone, then waves of 4. This applies to every `fetchAllFromTable` caller repo-wide, including the Inngest jobs and MRP.
2. **`itemStockQuantities` (item × location) was read in full on mount, again on a 10-minute interval, and again on every realtime stock change** — to decorate the Item combobox with an on-hand number, its only consumer in the app. A shop-floor tab left open re-downloaded the table per burst of movements. It is now a location-scoped, zero-filtered API route fetched only when a picker is mounted, refreshed by invalidation.
3. **Most realtime subscriptions had no server-side filter.** Only `itemStockQuantities` carried one; `item`, `customer`, `supplier`, `employee` and 9 `useRealtime` call sites subscribed unfiltered, so every tenant's row changes were pushed to every connected client and discarded in JS. All are now `companyId=eq.<id>`.

**Scope note, stated plainly.** The five remaining tables (`item`, `supplier`, `customer`, `employees`, `itemSupersession`) are still downloaded in full on every page load — that is unchanged and intentional for this PR. An earlier draft of this description blamed a 686-request/3.3 MB trace on table paging; that was not established. In that trace each table appears as a *single* request (`Content-Range: 0-76/77`), and the bulk of the count is PostHog `/e/` and `/s/` plus `__manifest` prefetches, none of which this PR touches. The paging fix only bites once a table exceeds 1000 rows.

Not a visual change. Before/after on a page load, captured at the network layer:

| | Before | After |
|---|---|---|
| Tables read on page load | 6 | 5 (`itemStockQuantities` gone) |
| `Prefer: count=exact` per page | yes | no |
| Paging past page 1 | serial | 4 concurrent |
| `itemStockQuantities` full reads per session | 1 + every 10 min + every stock movement | 0 |
| Unfiltered realtime subscriptions | 4 in the shell + 9 `useRealtime` call sites | 0 |

The five remaining table reads are unchanged — one request each, `companyId`-scoped.

The only user-visible change: an item with a stored `0` on-hand row used to render `0 EA` in the picker; it now renders no badge, the same as an item with no row at all.

## How should this be tested?

No new environment variables, no migration. A seeded dev company (`pnpm db:seed:dev`) with at least one item that has stock is enough.

**Automated** — `pnpm --filter @carbon/database test` (6 new tests in `packages/database/src/utils.test.ts` pin the pager: ordering across pages, the exact-multiple boundary, concurrency, error propagation instead of a partial read, and the `MAX_PAGES` backstop when a server ignores `Range`). Also green: `pnpm run lint`, and scoped typechecks for `erp`, `mes`, `@carbon/database`, `@carbon/jobs`, `@carbon/ee`.

**Manual (all verified against a local stack)**

1. Load `/x` with DevTools open, filtered to the PostgREST origin. Expect exactly five hydrate reads (`item`, `supplier`, `customer`, `employees`, `itemSupersession`), one request each, every URL carrying `companyId=eq.<id>`, **no `itemStockQuantities`**, and **no `Prefer: count=exact`** request header on any of them.
2. Open any form with an Item picker (e.g. Part → Planning → Add Predecessor). Expect one `GET /api/items/quantities/<locationId>`, on-hand badges matching that response exactly (`BRG-6201 | 24 EA`, `FST-M4-TI | 500 EA`), and no badge on items with no stock. Navigate away and back — no second request inside the 2-minute `staleTime`.
3. Inspect the realtime WebSocket `phx_join` frames: `item`, `customer`, `supplier`, `employee`, `itemStockQuantities` and `part` should all carry `filter: "companyId=eq.<id>"`.
4. Post a stock movement in a second tab and confirm the badge refreshes on next open, with no full-table read in the trace.

Not reproducible on a small seed: multi-page paging needs a tenant with >1000 items. That path is covered by the unit tests above.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Item pickers now display on-hand quantities for a selected location or across all locations.
  - Inventory quantity data is available through the item quantities API and read-only integration tool.

- **Bug Fixes**
  - Inventory badges remain stable when quantity data cannot be retrieved.
  - Realtime inventory updates are scoped to the active company and refresh matching quantities automatically.
  - Improved pagination reliability and performance for large data sets.

- **Documentation**
  - Clarified database pagination conventions and query requirements.


sidwebworks:
Both review findings were real and are fixed in 3904733671.

**1. `fetchAllRecords` — errors after the terminal page.** Correct: the pre-scan checked every result in a wave *before* finding the short page that ends the read, so a 1500-row table whose speculative page 3 failed returned `data: null` with every row already in hand. Pages past the terminal one are issued before we know where the data stops, and a read past the end can legitimately fail (PostgREST answers an out-of-range `Range` with 416). Now processed in page order, returning at the first short page. Added the suggested 1500-row test — it fails on the previous code and passes on this one.

**2. `useItemQuantities` — copied result, never observed.** Also correct, and the badge was stale on a mounted picker. It is now a `useQuery` on the same key; `root.tsx:275-297` shares one `QueryClient` between `window.clientCache` and the provider, so the imperative and hook paths hit the same entry.

**A third bug found while fixing that one.** The invalidation predicate also matched `queryKey[1]` against the companyId — and that term comes from `getCompanyId()`, which parses `document.cookie`. Every cookie in this app is httpOnly, so it is always `"null"` and the predicate matched nothing; the invalidation was dead code even once the query was observed. It now matches the key prefix alone. Keys for another company only exist after a company switch, which reloads the page and empties this in-memory cache anyway.

Verified against a local stack with a picker open: seeding the cache entry moved the badge `24 EA` → `999 EA` (it observes), then invalidating with the shipped predicate refetched it back to `24 EA` with no remount (the invalidation reaches it).

Also merged `main` in — the only conflict was the generated MCP tool manifest digest, resolved by regenerating (`pnpm run generate:mcp`, 1578 tools). Gates re-run green after the merge: lint, 7/7 `@carbon/database` tests, typecheck on `erp`, `mes`, `@carbon/database`, `@carbon/jobs`, `@carbon/ee`.

Not acting on the docstring-coverage warning: the functions it counts are small React hooks and a test double, and the two non-obvious ones (`fetchAllRecords`, `useItemQuantities`) already carry docstrings explaining *why* rather than restating the signature.

sidwebworks:
Correcting my own description — it claimed more than this PR delivers, and I have edited it.

**The shell still downloads whole tables.** `item`, `supplier`, `customer`, `employees` and `itemSupersession` are still read in full, from the browser, on every page load. That is unchanged here and needs server-side search comboboxes to fix properly, which is a separate piece of work. Measured on a dev tenant just now: 34 / 7 / 4 / 1 / 0 rows, ~10.7 KB total — small only because the tenant is small.

**The 686-request framing was not established.** I attributed that trace to table paging without proving the split. In the trace each table appears as a *single* request (`Content-Range: 0-76/77`), and the bulk of the count is PostHog `/e/` and `/s/` plus `__manifest` prefetches — none of which this PR touches. The paging fix only bites once a table exceeds 1000 rows.

What this PR does still fix, and I would defend:

- `count: "exact"` removed from every page of every `fetchAllFromTable` call repo-wide (Inngest jobs and MRP included) — that was a `COUNT(*) OVER ()` per page across the full filtered set, for a count nothing read.
- `itemStockQuantities` no longer downloaded at all: no mount read, no 10-minute poll, and no full re-read per burst of stock movements on a shop-floor tab left open.
- Unfiltered realtime subscriptions — every tenant's row changes were being pushed to every connected client and discarded in JS.

Reviewers: judge it on those three, not on the request count I originally quoted.

PR #1670, opened 2026-09-18
feat(items): annotate slides from the Bill of Process preview

sidwebworks:
## What does this PR do?

Adds the ability to open the "Annotate image" editor (drop/edit numbered pins) directly from the read-only Bill of Process **Preview** tab, not just the small editor thumbnails — the large preview image is a much better surface for placing pins. An "Annotate" button appears on the selected image slide (hidden when the method is read-only), opens the existing `SlideAnnotator`, and persists via the same `methodOperationStepSlide` upsert route the editor uses, so the pins immediately show in both the preview overlay and the editor. No route/schema changes.

On the Preview tab, an image slide now shows an "Annotate" button; clicking it opens the pin editor and saved pins render as the read-only overlay.

## How should this be tested?

Open a Part → make method → an operation whose step has an image slide → the **Preview** tab. Click "Annotate" on the slide, drop/move pins, Save; confirm the pins persist and appear both in the preview overlay and the editor tab. Confirm the button is hidden for a read-only (Active/locked) method, and that model slides show no Annotate button.

## Note

Stacked on `sid/task-preview-first-slide-only` (PR #1669), which introduces the multi-slide preview this builds on. Retarget to `main` once that merges.

## Checklist

- My code follows the style guidelines of this project (biome + scoped typecheck pass).


sidwebworks:
Closing — annotations are already shown read-only in the preview via #1669 (SlidePinOverlay). Editing annotations from the preview is not wanted.

PR #1679, opened 2026-09-20
fix(mcp): inject the acting user, surface edge-function errors, accept the documented call shape

BettrCallRaul:
## The bug

Creating a batch over MCP or an API key could never succeed, whatever arguments the caller sent:

```
production_createJobOperationBatch { jobOperationIds, locationId }
→ Database error: {"name":"FunctionsHttpError","context":{}}
```

The edge-runtime log says what the caller could not:

```
Error in batch-operations: ZodError: path ["userId"] — Required
```

## Root cause

A service takes the acting user one of two ways: as a positional `userId` param, which the dispatcher fills from context (`dispatch.server.ts`), or as a **field of its payload object** — the shape every edge-function wrapper uses:

```ts
export async function createJobOperationBatch(
  client: SupabaseClient<Database>,
  args: { jobOperationIds: string[]; locationId: string; …; companyId: string; userId: string }
) {
  return client.functions.invoke("batch-operations", { body: { type: "create", ...args } });
}
```

`CONTEXT_PARAMS` strips `userId` from the published schema on the assumption something supplies it, but `AuthField` had no `userId`, so nothing did. **95 tools** reached their service with no acting user.

Where the payload crosses into an edge function the failure is hard — the function validates and rejects. Where it stays in Postgres it is quiet: `notificationPreference` upserts on `userId,companyId,channel,topic`, so the row could not land correctly.

None of it was visible through the app, which passes `userId` explicitly. It reached only API-key and MCP callers.

This is not a behaviour change so much as an unimplemented one — `.claude/rules/mcp-tools-reference.md` already states the contract:

> `companyId`/`userId` come from the auth context and are injected server-side — never trusted from tool arguments.

The gap was noticed once before, in review on #1320, and handled by blocklisting the two tools that hit it ("their args require a userId the MCP executor cannot inject (AuthField has no such payload field)").

## What changed

**1. Inject the acting user** — `userId` joins `AuthField`; the generator derives it from the signature so the strip and the injection stay in step; the dispatcher stamps it, overwriting any caller-supplied value the way it already does for `createdBy` (the field is not in the schema, so a caller sending one is forging attribution).

**2. Surface edge-function errors** — the message lives in an unread `Response` on `error.context`, which `JSON.stringify` empties, so every failure looked identical. It now unwraps the body with the helper the routes already use. Every other Supabase error keeps the byte-exact legacy text the parity tests pin.

**3. Accept the documented call shape** — the published instructions teach `arguments: { args: {...} }`, but only operations that declare an `args` object take it; for the rest the envelope IS the payload, so every declared field arrived undefined. An agent following the server's own documentation failed on its first required-parameter call. A lone envelope is now unwrapped before validation, and the instructions show the flat shape. List reads masked this entirely — their parameters are all optional.

## Verification

Driven through a real MCP client against a local stack, API-key auth.

| | before | after |
|---|---|---|
| create, flat args | `{"name":"FunctionsHttpError","context":{}}` | `BAT000001`, `createdBy` = the key's user |
| create, on a non-batchable process | same opaque string | `The process is not batchable` |
| create, on an already-batched operation | same opaque string | `Operation jo_9JAq… is already in a batch` |
| create, documented `{args:{…}}` shape | `jobOperationIds: expected array, received undefined` | same as flat |
| release → unrelease round trip | not reachable | `{"success":true}`, status back to Planned, `updatedBy` stamped |

`apps/erp`: 112 files, 1306 tests passing. `tsgo --noEmit` clean.

## Guard

`apps/erp/test/mcp-tool-auth-injection.test.ts` fails if a service payload declares `userId` without the manifest marking it for injection — the check that would have caught this when job operation batching shipped. Confirmed it bites by stripping the field from the batch tool's entry.

## Notes for review

- **One parity expectation changes**: `account_upsertNotificationPreference` now receives the `userId` its payload declares. That is the bug being fixed, not a regression.
- **`account_upsertUserAttributeValue` / `deleteUserAttributeValue`** take `userId` as the *subject*, not the actor, so injection pins them to the caller's own attributes. That matches the app's only usage; setting another user's attribute over MCP was never possible, since the field is stripped from the schema.
- **User-scoped RLS is unaffected and still bites.** `notificationPreference`'s policies are `"userId" = auth.uid()`, and an API key is not a Supabase JWT, so the write now reaches the database and is honestly rejected instead of failing on a missing field. OAuth clients get `getUserScopedClient(userId)`, so `auth.uid()` is set and these tools work for them — previously they were broken for everyone. Left reachable rather than extended onto the blocklist, but happy to sweep the 95 for user-scoped tables if that is preferred.
- **`production_returnPickedRemaindersForOperation` / `ForJob`** were blocklisted only because of this gap and could now be unblocked. Left alone here — separate decision, separate PR.

## Summary by CodeRabbit

- **New Features**
  - MCP tools now accept parameters directly in the `arguments` object.
  - Legacy `{ args: {...} }` requests remain supported when compatible with the tool schema.
  - Authentication context now supplies the acting user ID where required.

- **Bug Fixes**
  - Corrected handling of wrapped and flat MCP tool arguments during dispatch.
  - Database and edge-function failures now return consistent classified messages without exposing sensitive details.

- **Documentation**
  - Updated MCP usage guidance to explain supported parameter formats and tool schemas.


apps/erp/app/routes/api+/v1+/lib/args-envelope.ts
```diff
@@ -0,0 +1,17 @@
+export function unwrapArgsEnvelope(
+  meta: { schema?: unknown },
+  args?: Record<string, unknown>
+): Record<string, unknown> | undefined {
+  if (!args) return args;
+  const declaresArgs = Boolean(
+    (meta.schema as { properties?: Record<string, unknown> } | undefined)
```
naveenkash: Can we make it typed?

apps/erp/app/routes/api+/v1+/lib/call.server.ts
```diff
@@ -19,10 +26,20 @@ import {
   type StandardIssue
 } from "./validation-issues";
 
+const logger = getLogger("erp", "api", "call-operation");
+
 export type CallResult =
   | { success: true; data: unknown; count?: number }
   | { success: false; error: string; errorKind: "database" | "execution" };
 
+async function edgeFunctionMessage(error: unknown): Promise<string | null> {
```
naveenkash: Can we make it typed?

naveenkash (approved):

